PR #92, opened 2014-09-09
Add libboost-filesystem-dev to dependencies

gravitystorm:
Curiously it fails with a misleading error (configure: error: Could not link against boost_date_time-mt !) but that's not too important.


zerebubuth:
Thanks!

PR #98, opened 2015-07-11
Create getto

curl6425:


zerebubuth:
I don't understand why I would want to create an empty file called "getto". Please could you provide a description for the change, and the reason why you think it would improve the project?

PR #104, opened 2016-02-25
Add headers to cors preflight

### LINKED ISSUE #102: CORS problem - fails to indicate all headers are allowed
An attempt to run a small OpenLayers 2 example with a vector layer populated directly from the OSM API like so

<pre>
var layer = new OpenLayers.Layer.Vector("OSM Data", {
    strategies: [new OpenLayers.Strategy.Fixed()],
    protocol: new OpenLayers.Protocol.HTTP({
        url: "http://api.openstreetmap.org/api/0.6/map?bbox=8.3841,49.0115,8.3873,49.0131",
        format: new OpenLayers.Format.OSM()
    }), 
    projection: new OpenLayers.Projection("EPSG:4326")
}); 
</pre>

fails with a browser error message of 

> Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at http://api.openstreetmap.org/api/0.6/map?bbox=8.3841,49.0115,8.3873,49.0131. (Reason: missing token 'x-requested-with' in CORS header 'Access-Control-Allow-Headers' from CORS preflight channel).

This has its roots in OpenLayers automatically adding an "X-Requested-With" header to the AJAX request.

CGIMap seems to allow all headers, so this should not fail, but CGIMap fails to properly indicate to the browser in the "preflight" phase that it will allow all headers, hence the browser refuses to execute the actual request.

CGIMap needs to respond with a proper Access-Control-Allow-Headers header.

zerebubuth:
Fixes #102.

@woodpeck - how does that look?

### DIFF
--- a/src/process_request.cpp
+++ b/src/process_request.cpp
@@ -189,7 +189,7 @@ boost::tuple<string, size_t> process_options_request(request &req) {
   const char *origin = req.get_param("HTTP_ORIGIN");
   const char *method = req.get_param("HTTP_ACCESS_CONTROL_REQUEST_METHOD");
 
-  if (origin &&
+  if (origin && method &&
       (strcasecmp(method, "GET") == 0 || strcasecmp(method, "HEAD") == 0)) {
 
     // write the response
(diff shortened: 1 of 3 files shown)


woodpeck:
I ran my original tests against the live OSM API and can't immediately run it against this version but it certainly looks good!

PR #106, opened 2016-04-14
fixed a bug that in daemon mode, legacy "port" option fallback not recognized.

frankxu2004:
added the "port" option for daemon mode in `get_options` function.


zerebubuth:
That's great, thanks for fixing it!

PR #107, opened 2016-04-16
Allow negative changeset IDs

### LINKED ISSUE #105: ERROR:Could not convert string to unsigned integer
After dumping the data dump using the osmosis tool int the database on my server, all queries to the API recieved error message: 
`<osm version="0.6" generator="CGImap 0.4.0 (30216 galaxy)" copyright="OpenStreetMap and contributors" attribution="http://www.openstreetmap.org/copyright" license="http://opendatacommons.org/licenses/odbl/1-0/">
<bounds minlat="31.2310779" minlon="121.5241742" maxlat="31.2369124" maxlon="121.5585065" />
<error>Could not convert string to unsigned integer: '-1'</error>
</osm>
`
I am not really sure that happened, after a quick search I thought that  I have been fixed previously due to anomynous users handling.

zerebubuth:
This PR separates the different types of IDs (node/way/relation, user, changeset, tile) into different typedefs and makes the changeset ID signed. This should enable cgimap to deal with databases where Osmosis has made a "-1" changeset, which it seems to do when reading PBF files.

@frankxu2004 could you check out this branch and see if it fixes #105, please?

### DIFF
--- a/include/cgimap/api06/handler_utils.hpp
+++ b/include/cgimap/api06/handler_utils.hpp
@@ -8,8 +8,8 @@
 
 namespace api06 {
 
-std::vector<osm_id_t> parse_id_list_params(request &req,
-                                           const std::string &param_name);
+std::vector<osm_nwr_id_t> parse_id_list_params(request &req,
+                                               const std::string &param_name);
 }
 
 #endif /* API06_HANDLER_UTILS_HPP */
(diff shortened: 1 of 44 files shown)


frankxu2004:
Yes I have tested and now with that negative ID handle, it is totally fine to respond correct data. It works for me now. Thanks a lot!

zerebubuth:
Great, thanks!

Issue #96, opened 2015-06-23
Support for verifying OAuth signatures


zerebubuth:
Some API calls are signed with [OAuth](http://oauth.net/) signatures to authenticate the user. This is done using the `oauth` gem in the website (apparently we support only 1.0 & 1.0a, not 2), but there is no way of doing this at present in cgimap.

Support for verifying signatures would be the first step toward per-user rate limiting, rather than per-IP, as discussed [on the operations tracker](https://github.com/openstreetmap/operations/issues/36).

zerebubuth:
Fixed by #110.

PR #110, opened 2016-07-18
OAuth-based rate limiting

zerebubuth:
This adds OAuth support and uses it to influence rate limiting behaviour. When the incoming request has a valid OAuth signature, then the key used to look up rate limits is formed using the user's ID rather than the source IP address. This means that several users behind a firewall can all have individual byte counters and they won't get rate limited any sooner than they would at home.

Adding OAuth signatures to GET requests is, I think, not yet common behaviour amongst editors. Most will have support for PUTs, so it would probably be easy enough to add for GETs too.

This should make a significant difference to openstreetmap/operations#36, perhaps even solving it. If not, further steps to take might be to integrate explicit configurable whitelists into cgimap.

@Firefishy, @tomhughes could you take a look at the code, please, and let me know if you think it looks okay?


include/cgimap/http.hpp
```diff
@@ -115,17 +116,32 @@ class gone : public exception {
...
+std::string urlencode(const std::string &s);
+
+/**
  * Parses a query string into a key-value map.
  *
  * The string should already have been url-decoded (i.e: no %-encoded
  * chars remain).
  */
-std::map<std::string, std::string> parse_params(const std::string &p);
+std::vector<std::pair<std::string, std::string> > parse_params(const std::string &p);
```
tomhughes: What's the reason for changing from a map? If it's to allow duplicates is multimap not the way to go?

include/cgimap/http.hpp
```diff
@@ -115,17 +116,32 @@ class gone : public exception {
...
 
 /**
+ * Encodes a string to canonical url-encoding.
+ *
+ * This is compatible with the OAuth 1.0a definition of url-encoding.
+ */
+std::string urlencode(const std::string &s);
+
+/**
  * Parses a query string into a key-value map.
```
pnorman: It's no longer a map

include/cgimap/http.hpp
```diff
@@ -3,6 +3,7 @@
 
 #include <string>
 #include <map>
```
pnorman: No longer used

zerebubuth:
Note: Also fixes #96.

include/cgimap/http.hpp
zerebubuth: That was to handle duplicates in commit 4dfa54cdb42eb48b3258b98aa4421d255acf08f6. One of the [OAuth examples](https://tools.ietf.org/html/rfc5849#section-3.4.1.3.1) has duplicates coming from a mixture of GET query and POST form-encoded parameters, and the code wasn't handling that correctly. It seemed easier to remove the assumption that query parameters had unique keys throughout the code than to convert back and forth between data structures.

As for `multimap` vs `vector<pair>`... I just figured the latter was simpler to deal with, but perhaps it would be a nice code cleanup later to move to `multimap`.

PR #111, opened 2016-08-21
Add support for changesets

zerebubuth:
This adds support for changesets and changeset discussions to cgimap.

@tomhughes, @pnorman your comments would be most appreciated. Especially if you think there's any test cases which need to be added to cover corner cases or ensure compatibility in behaviour with the rails code?


src/backend/apidb/writeable_pgsql_selection.cpp
```diff
@@ -68,6 +71,27 @@ template <> struct string_traits<vector<tile_id_t> > {
...
+  static void from_string(const char[], vector<osm_changeset_id_t> &) {}
+  static std::string to_string(const vector<osm_changeset_id_t> &ids) {
+    stringstream ostr;
+    ostr << "{";
+    std::copy(ids.begin(), ids.end(),
+              infix_ostream_iterator<osm_changeset_id_t>(ostr, ","));
+    ostr << "}";
+    return ostr.str();
+  }
+};
```
pnorman: We now have basically the same code repeated for osm_nwr_id_t, tile_id_t, and osm_changeset_id_t. Is there any reasonable way to deduplicate?

src/time.cpp
```diff
@@ -0,0 +1,53 @@
+#include "cgimap/time.hpp"
+#include <boost/date_time/time_parsing.hpp>
+#include <stdexcept>
+#include <sstream>
+#include <ctype.h>
+
+boost::posix_time::ptime parse_time(const std::string &s) {
```
pnorman: Are there no functions in the libraries which can do this?

pnorman:
Right now MAX_NODES is defined in only `src/api06/map_handler.cpp` but this adds a few definitions of 50k, and it's not clear to me what all of them are for.

With openstreetmap/openstreetmap-website#1259 we have a PR to change the max changeset upload size in the API, and it's been different in the past with an off-by-one error. Will either of these cause problems?

pnorman:
Do we have a plan for how to test this, ideally with a canary deployment? Is there a better place to discuss that?

src/backend/apidb/writeable_pgsql_selection.cpp
zerebubuth: Not as reasonable as I'd have liked (i.e: hiding it away in a `.cpp` file), but I've reduced it to a macro in fb6ad22 which at least means there's only one copy of the code in the human-readable sense.

src/time.cpp
zerebubuth: I can't remember why I did it the long way before. Perhaps because it was simple and fixed-format enough to not bother with calling some other function. I've made it call a Boost function in 701caa6. I don't know if it's any clearer now, since the Boost function is pretty much undocumented (and calls other parsing functions).

Apart from testing, it only seems to be called to parse changeset `closed_at` times from the database, which are formatted in a known way, so it shouldn't be a problem.

zerebubuth:
> Right now MAX_NODES is defined in only `src/api06/map_handler.cpp` but this adds a few definitions of 50k, and it's not clear to me what all of them are for.

Well spotted! It seems they're completely unused and I've removed them in 0697e91.

> we have a PR to change the max changeset upload size in the API, and it's been different in the past with an off-by-one error. Will either of these cause problems?

Yes. The only place in the code that the max changeset size is used is [when figuring out if a changeset is closed](https://github.com/zerebubuth/openstreetmap-cgimap/blob/0697e91b69cf8585f7399e1b5a08f3fc28579570/src/output_formatter.cpp#L58-L61), and it really ought to be configurable. I'll think about how to add that with minimal pain.

As for the off-by-one errors, I think the best way to test that is with a set of tests for all the different corner cases. Help with adding more tests would be greatly appreciated.

> Do we have a plan for how to test this, ideally with a canary deployment? Is there a better place to discuss that?

My plan was to have as comprehensive a suite of tests as possible. We can certainly manually do a canary deployment, but I'd prefer to be confident ahead of time that cgimap is doing the right thing because it's passing the tests.

pnorman:
> My plan was to have as comprehensive a suite of tests as possible. We can certainly manually do a canary deployment, but I'd prefer to be confident ahead of time that cgimap is doing the right thing because it's passing the tests.

I think we have the HTTP status, headers, and other similar things in agreement between the rails port and cgimap, so it's just changeset logic we should need to worry about, but comprehensive tests aren't enough.

Last time with the relation full call we had a comprehensive set of tests and I had tested it for performance but when it was pointed against ramoth the combination of the database statistics, postgresql version and statistics tuning lead to it doing sequential scans.

zerebubuth:
> I think we have the HTTP status, headers, and other similar things in agreement between the rails port and cgimap, so it's just changeset logic we should need to worry about, but comprehensive tests aren't enough.

Agreed, and I think we can try to do a canary deployment. Comprehensive tests are necessary, but not sufficient. I don't think the unit tests are anything like comprehensive at the moment. Given you're asking about `MAX_NODES` off-by-one errors, neither do you.

pnorman:
> I don't think the unit tests are anything like comprehensive at the moment. Given you're asking about `MAX_NODES` off-by-one errors, neither do you.

Well, that was more about the question of how we're handling changesets with > 50k changes.

PR #112, opened 2016-10-03
Space separate parameters in oauth connection string

tomhughes:


zerebubuth:
Ooops. Well spotted, thanks! :smile:

PR #114, opened 2016-10-12
Install all packages using apt addon

tomhughes:
Lets be consistent about how we install packages for travis...


zerebubuth:
Good point, thanks!

PR #115, opened 2016-10-12
Don't require sudo for travis

pnorman:
This should reduce the time waiting for a build machine on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: required
+sudo: false
 dist: trusty
 language: cpp
 compiler: gcc


pnorman:
libpqxx-dev isn't whitelisted, I'll look into getting it added to Travis.

zerebubuth:
It seems to have a problem with `libboost-locale-dev` as well:

```
E: Unable to locate package libboost-locale-dev
```

Does that also need to be whitelisted? Seems like the other `libboost-*dev` are.

pnorman:
Can you rerun travis and we'll see what errors we get.

zerebubuth:
[Re-ran](https://travis-ci.org/zerebubuth/openstreetmap-cgimap/builds/167113797), but got the same error:

```
$ sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install libboost-dev libboost-date-time-dev libboost-filesystem-dev libboost-locale-dev libboost-program-options-dev libboost-regex-dev libboost-system-dev libcrypto++-dev libfcgi-dev libmemcached-dev libpqxx-dev libxml2-dev
Reading package lists...
Building dependency tree...
Reading state information...
Package libpqxx-dev is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  libpqxx3-dev:i386 libpqxx3-dev

E: Unable to locate package libboost-locale-dev
E: Package 'libpqxx-dev' has no installation candidate
```

It looks like the version of libpqxx (>=4) we need isn't present, and neither is libboost-locale-dev.

PR #117, opened 2016-10-30
Add API calls for specific historic versions of objects

woodpeck:
This pull requests adds support (and tests) for requests of the form /.../{node|way|relation}/[id]/[version]


zerebubuth:
Thanks!

PR #118, opened 2016-11-03
Reject oauth nonces over a day old

tomhughes:
This (along with the corresponding rails commit) will allow us to expire old nonces rather than having to keep them around forever.


zerebubuth:
Thanks!

Issue #124, opened 2016-12-24
HTTP 400 error message explanations: HTTP header vs. body


mmd-osm:
Apologies, if this has been reported before.

I'm wondering a bit, why the error message `Error: You requested too many nodes (limit is 50000). Either request a smaller area, or use planet.osm` is sent back as HTTP header field "Error" (which appears to be somewhat non-standard).

I'd rather expect this kind of information as part of the HTTP body. As a nice side effect, the error message would also be shown in the browser.

```
*   Trying 2001:630:12:500:219:bbff:fe39:8aba...
*   Trying 193.63.75.99...
* Connected to api.openstreetmap.org (2001:630:12:500:219:bbff:fe39:8aba) port 80 (#0)
> GET /api/0.6/map?bbox=1.9885,48.7659,2.5107,48.9755 HTTP/1.1
> Host: api.openstreetmap.org
> Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
> Accept-Encoding: gzip, deflate
> Accept-Language: de,en-US;q=0.7,en;q=0.3
> Connection: keep-alive
> DNT: 1
> Referer: http://www.openstreetmap.org/export
> Upgrade-Insecure-Requests: 1
> User-Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:50.0) Gecko/20100101 Firefox/50.0
> 
< HTTP/1.1 400 Bad Request
< Date: Sat, 24 Dec 2016 11:45:30 GMT
< Server: Apache/2.4.18 (Ubuntu)
< Content-Disposition: attachment; filename="map.osm"
< Error: You requested too many nodes (limit is 50000). Either request a smaller area, or use planet.osm
< Cache-Control: no-cache
< Content-Length: 0
< Content-Type: text/html; charset=utf-8
< Connection: close
< 
* Closing connection 0
```

zerebubuth:
Thanks for pointing that out!

That's a regression and a bug. I've made an attempt at a fix in #125.

mmd-osm:
Great, thanks a lot! Is there anything else left to do or can this issue be closed then?

zerebubuth:
Yup, good point! Fixed in #125.

Issue #83, opened 2013-10-25
Support missing X-Error-Format request header


zerebubuth:
The X-Error-Format request header allows a client to request that any error reported by the client is both as an Error header and XML in the body of the response. If this is missing, or isn't "xml", then the response should contain the error as a "text/plain" string. CGImap doesn't.

See https://github.com/openstreetmap/openstreetmap-website/blob/master/app/controllers/application_controller.rb#L266 for the rails implementation.

zerebubuth:
Looks like @tomhughes fixed this in becfb427 as part of #3.

PR #125, opened 2017-01-15
Fix HTTP error messages

zerebubuth:
Thanks to @mmd-osm for pointing this out in #124.

It looks like this sort of worked as of #83, although returning HTML rather than plain text as the Rails code does. That was sort of fixed in 1d0e06f2bdcbb5efc5ec7908e79d5d4e65fd139f to match the behaviour of _most_ of the Rails errors: which is to print a blank string as the error. Unfortunately, the side effect was disabling display in the body of all messages.

This branch tries to fix this, first by printing the error messages in the response body and secondly by providing error messages for all the various `not_found` cases.

This is a departure from strict compatibility with the Rails code. However, most of the `not_found` messages that the Rails code generates are translations from `ActiveRecord::RecordNotFound`. If I remember correctly, at the time of writing it was difficult to tell exactly _what_ hadn't been found, although I see that [more information might be available now](http://api.rubyonrails.org/classes/ActiveRecord/RecordNotFound.html). I think the best thing is to provide a meaningful error message in cgimap, and I'll try to come up with a similar patch for the Rails code.

@tomhughes what do you think?

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -74,7 +74,7 @@ endif
 TESTS = test/map.testcore test/node.testcore test/anon.testcore test/way.testcore test/relation.testcore
 TESTS += test/empty.testcore test/way_full.testcore test/relation_full.testcore
 TESTS += test/test_parse_id_list test/test_oauth test/test_http test/test_parse_time
-TESTS += test/changesets.testcore
+TESTS += test/changesets.testcore test/message.testcore
 if ENABLE_EXPERIMENTAL
 TESTS += test/node_ways.testcore
 endif
--- a/src/api06/changeset_handler.cpp
+++ b/src/api06/changeset_handler.cpp
@@ -23,7 +23,9 @@ changeset_responder::changeset_responder(mime::type mt, osm_changeset_id_t id_,
   }
 
   if (sel->select_changesets(ids) == 0) {
-    throw http::not_found("");
+    std::ostringstream error;
+    error << "Changeset " << id << " was not found.";
+    throw http::not_found(error.str());
   }
 
   if (include_discussion) {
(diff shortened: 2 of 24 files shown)


tomhughes:
Does the rails code return anything in the body? I thought it was just the `Error` header unless you asked for XML errors, because it was odd to return a different sort of document on error...

tomhughes:
Looks like rails does also return a text body with the error so I guess this is fine.

Whether the person who contacted ops this week would have noticed the body any more than the error header is an open question ;-)

zerebubuth:
My reading of [`application_controller.rb:287`](https://github.com/openstreetmap/openstreetmap-website/blob/96c91757fc23023f2d56019252a9f89f2482266f/app/controllers/application_controller.rb#L287) is that it does, when a message is available. But it looks like [not found explicitly returns the empty string](https://github.com/openstreetmap/openstreetmap-website/blob/96c91757fc23023f2d56019252a9f89f2482266f/app/controllers/application_controller.rb#L318).

I suppose we could replace the body with `<script>alert("#{error_message}");</script>`, but I figure that would get annoying very quickly :wink:

zerebubuth:
Upon further investigation, it looks like the extra details in `ActiveRecord::RecordNotFound` are only in [Rails 5.0.1 or later](https://github.com/rails/rails/commit/d763956ed95bce581f12151a5ed4df12bcefdeee). And Rails 5 is still burning hot from the oven...

In the meantime, I'll just leave this how it is for the moment. One small fix: looks like Rails doesn't support the XML error stuff for 404s, so I'll make it match that.

tomhughes:
Is that sort of bug-for-bug compatibility really sensible? Shouldn't we just fix the rails code? Why does it special case record not found?

tomhughes:
Looking at it I guess the logic is that a 404 has meaning on it's own. Quite why it is giving text/html for a `:text => ""` render is another question...

PR #126, opened 2017-02-08
Add documentation about lighttpd usage

jronak:
Added instructions for installing and running cgimap with lighttpd

### DIFF
--- a/README
+++ b/README
@@ -81,12 +81,21 @@ directory.
 Setup
 -----
 
-A sample lighttd.conf file is provided, which can be used to test. To test
-CGImap with lighttpd use the supplied config file with lighttpd and run cgimap
-on the command line with a command like
+To run CGImap binary, use the command:
 
     ./openstreetmap-cgimap --socket :54321 --backend apidb \
       --dbname openstreetmap
+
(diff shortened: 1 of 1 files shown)


README
```diff
@@ -82,8 +82,15 @@ Setup
...
-on the command line with a command like
+CGImap with lighttpd you will need to install lighttpd:
+
+	sudo apt-get install lighttpd
+
+Edit the supplied lighttpd.config file to include your CGImap path and run it with the lighttpd like
+
+	/usr/sbin/lighttpd -f lighttpd.conf
+
+Run CGImap on the command line with a command like
```
pnorman: Didn't we conclude that lighttpd handles starting cgimap and if you're using the config file you don't need to also start the binary?

jronak: Yes, FastCGI handles starting the binary, but I thought we are starting the binary for debugging?

pnorman: If lighttpd starts the binary then there's no purpose running a second copy

pnorman: Note: we still need to document how to start the binary, just make it clear you don't do that when running it through lighttpd with the supplied config file

jronak: Updated PR, please review.

zerebubuth:
Thanks!

PR #127, opened 2017-02-13
Add unknown route handler in lighttpd configuration

### LINKED ISSUE #52: Reject unknown API calls
A request to `/api/0.6/way/#id/foo` functions the same as a request to `/api/0.6/way/#id`.

This caused me some confusion when trying to figure out why I wasn't getting back ways and nodes (I was on a branch without full handlers moved out of experimental).

Not a huge issue because the HTTP server config directing to cgimap probably won't normally direct requests it can't handle, but a clearer error would be nice.

jronak:
Fixes #52, routes the unknown call to the cgimap to display a verbose message. I haven't included the code to return an error message as it is being handled in #125.

### DIFF
--- a/lighttpd.conf
+++ b/lighttpd.conf
@@ -7,6 +7,9 @@ server.port = 31337
 # selecting modules
 server.modules = ( "mod_access", "mod_rewrite", "mod_fastcgi" )
 
+# handling unknown routes
+server.error-handler-404   = "/dispatch.map"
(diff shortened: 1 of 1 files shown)


zerebubuth:
Hi! This is great, thanks!

I think there's still some work to be done on #52, though. I'm about to add some detail there about what remains to be done.

PR #128, opened 2017-02-16
Fix reject unknown api calls with 404

### LINKED ISSUE #52: Reject unknown API calls

jronak:
Fixes #52. 
Solved it by checking if the request path has been matched/iterated completely. I think we can use the existing router with this fix for now until we have a new router.

Should I add the test cases as shown in 52-unknown-api-calls?

Let me know if I have to fix my code style/syntax.

Thanks!

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -74,7 +74,7 @@ endif
 TESTS = test/map.testcore test/node.testcore test/anon.testcore test/way.testcore test/relation.testcore
 TESTS += test/empty.testcore test/way_full.testcore test/relation_full.testcore
 TESTS += test/test_parse_id_list test/test_oauth test/test_http test/test_parse_time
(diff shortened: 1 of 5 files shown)


pnorman:
> Should I add the test cases as shown in 52-unknown-api-calls?

Yes, merge that git branch in and make sure it's passing

jronak:
Merged, the test passes.

zerebubuth:
Awesome! Thank you!

PR #129, opened 2017-02-17
Add support for history calls

zerebubuth:
This includes:

* Node, way and relation history calls, e.g: `api/0.6/node/#{id}/history`
* Node, way and relation version calls, e.g: `api/0.6/node/#{id}/#{version}`
* Node, way and relation multi-selection calls, e.g: `api/0.6/nodes?nodes=#{id}v#{version},...` as discussed in https://github.com/openstreetmap/openstreetmap-website/pull/1189

This is a _huge_ patch and I can't ask anyone to go through it with a fine-toothed comb. It would be really helpful if someone has some time to look through the test cases and suggest any that I might have missed or would be good to have coverage for.

@pnorman, @tomhughes, I would really appreciate your feedback.


test/redactions.testcore/node_version_moderator.case
```diff
@@ -0,0 +1,17 @@
+# node/version call for a regular user should behave like the redacted version
+# does not exist.
```
tomhughes: Comment is wrong - this is testing what happens for a moderator.

test/redactions.testcore/relation_version_moderator.case
```diff
@@ -0,0 +1,19 @@
+# relation/version call for a regular user should behave like the redacted version
```
tomhughes: This is the moderator test...

test/redactions.testcore/way_version_moderator.case
```diff
@@ -0,0 +1,19 @@
+# way/version call for a regular user should behave like the redacted version
+# does not exist. NOTE: OAuth stuff generated with http://lti.tools/oauth/
```
tomhughes: And again ;-)

test/relation.testcore/relation_history_404.case
```diff
@@ -0,0 +1,6 @@
+# relation/history call returns all versions, even deleted ones, in the request
```
tomhughes: This is testing a non-existent relation.

test/way.testcore/way_history_404.case
```diff
@@ -0,0 +1,6 @@
+# way/history call returns all versions, even deleted ones, in the request
```
tomhughes: Actually testing for missing object...

src/backend/apidb/oauth_store.cpp
```diff
@@ -160,3 +160,9 @@ oauth_store::get_user_id_for_token(const std::string &token_id) {
     return boost::none;
   }
 }
+
+std::set<osm_user_role_t>
+oauth_store::get_roles_for_user(osm_user_id_t id) {
+  // TODO: implement!
+  return std::set<osm_user_role_t>();
```
zerebubuth: D'oh! Thanks for spotting that.

I've added an implementation now, and a test for it as well.

zerebubuth:
@tomhughes thanks very much for the comments! I've attempted to fix all of those. If you have more feedback, it would be greatly appreciated.

Issue #123, opened 2016-12-18
Remove per-element tag queries


pnorman:
#122 does raise a valid issue, even if buried and unclear.

Currently each object in the output requires a SQL query to fetch its tags. I had looked at this in GSoC and noted it as a performance problem, but lower priority than other work.

In the examples below I'm just using nodes, but they'd apply to other object types too.

My thought at the time was to have tags returned at the time of selection, in a query that returns the tags as a column, either of type hstore, an array of alternating values, or two arrays. I'm inclined towards two arrays because I believe that to be easiest.

This would be something like
```sql
select n.id, ..., array_agg(k) keys, array_agg(v) vals
  FROM current_nodes n
    JOIN current_node_tags t ON (n.id=t.node_id)
  WHERE ...
  GROUP BY n.id -- primary key, so should work instead of listing all node columns
```

pnorman:
Fixed for nodes in #130, but other object types remain

jronak:
@pnorman #130 fixed for all the types, the title was incorrect. Please verify

zerebubuth:
Yup, #130 fixes it for all object types - including changesets.

I think the next steps are to do similar things for the way nodes and relation members, then extend that to the `readonly_apidb_backend`. And, after I merge #129, extend to the history calls as well (because they have a composite primary key on `id, version` they need similar, but slightly different, queries).

At some point I also want to see whether `writeable_apidb_backend` and `readonly_apidb_backend` can be refactored to share more code. At the moment, they share very little, but I feel like they ought to be able to share a great deal more since the table schema is identical.

jronak:
@zerebubuth I think this issue can be closed now

zerebubuth:
Yup, good point.

PR #130, opened 2017-02-19
Fix per-node separate tag query

jronak:
Draft PR for #123 

I have fixed the query only for write_node method, will fix other methods after initial review

Thanks

### DIFF
--- a/src/backend/apidb/writeable_pgsql_selection.cpp
+++ b/src/backend/apidb/writeable_pgsql_selection.cpp
@@ -249,7 +249,7 @@ void writeable_pgsql_selection::write_ways(output_formatter &formatter) {
        ++itr) {
     extract_elem(*itr, elem, cc);
     extract_nodes(w.prepared("extract_way_nds")(elem.id).exec(), nodes);
-    extract_tags(w.prepared("extract_way_tags")(elem.id).exec(), tags);
+    extract_tags(*itr, tags);
     formatter.write_way(elem, nodes, tags);
   }
 }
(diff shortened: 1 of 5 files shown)


zerebubuth:
Wow, thanks! This is wonderful.

The code looks good! Please feel free to go ahead with the other methods.

Just a few small comments:

* I'm not sure that `psql_array_to_vector` handles all the possible variations in [array syntax](https://www.postgresql.org/docs/current/static/arrays.html#ARRAYS-IO). Sadly, I didn't see a formal syntax there, but I think the input `{"}"}` or `{"\""}` would not be parsed correctly at the moment. Would it be possible to expose `psql_array_to_vector` in a header and write some unit tests for it, please?
* In `extract_tags`, the code loops over `keys.size()` and that should always be equal to `values.size()`, but it's possible that due to unexpected circumstances there might be differences. Could it either loop over `min(keys.size(), values.size())` or have an `assert(keys.size() == values.size())`, please? I think the latter would be preferable; although it would crash the process, it couldn't lead to undefined output.
* The code isn't formatted in the same way as [the rest of the project](https://github.com/zerebubuth/openstreetmap-cgimap#formatting). I don't mind if you don't want to reformat it - just let me know and I'll happily do that in the merge.

jronak:
Thanks for the feedback!

- Thanks for the array syntax page!, I shall rewrite it to handle cases as given in the syntax page. In fact, I tried many ways to split it such as boost::split but it didn't work well. 
- Thought of assertion, will add it.
- Sorry about that, actually I tried make clang-format but it gave 'no target clang-format' error. I will fix it in the next commit.

Thanks!

zerebubuth:
Thank you for the code!

Regarding the array syntax, it looks like a moderately complex grammar. In these situations, I might be tempted to use [Boost Spirit](http://www.boost.org/doc/libs/1_63_0/libs/spirit/doc/html/index.html)... and then I might regret that later! :smile: Boost Spirit is amazingly powerful, but I also find it amazingly infuriating when it isn't working how I expect it to.

One alternative may be to use `json_agg` instead of `array_agg`, and parse the JSON rather than the array syntax. Cgimap already uses [libyajl](https://lloyd.github.io/yajl/) if configured with `--enable-yajl`, so we could easily make that a required dependency. What do you think?

Regarding formatting, I'm not entirely sure that the `clang-format` stuff is working any more. Sorry about that! I'll have to try and get that working again. Until then, it's close enough to do a [1TBS](https://en.wikipedia.org/wiki/Indent_style#Variant:_1TBS_.28OTBS.29) with a 2 space indent and spaces after keywords like `if` and `while` but not after function calls. It's not a big deal.

jronak:
I would definitely love to try spirit! (I haven't checked it out before), but the array_syntax was pretty basic so just wrote a simple code to parse it (fairly simple). I think array will do the job now, no need to make libyajl a dependency now. I have tried my best to fix the syntax issues, pardon me if I have missed a fix. 

I have added the array to vector method in data_selection as it makes more sense there over write_apidb as the method can be used by other backends as well. I have written few test cases for the array to vector method which handle corner cases too (whatever I could think of :P ). 

Please review, 
thank you

src/backend/apidb/writeable_pgsql_selection.cpp
```diff
@@ -8,6 +8,7 @@
 #include <sstream>
 #include <list>
 #include <vector>
+#include <boost/algorithm/string/trim.hpp>
```
zerebubuth: We don't need this header in this file any more, do we?

jronak: Forgot to clean up, not needed.

zerebubuth:
That all looks great. Thank you!

PR #131, opened 2017-02-24
Fixes per element queries

jronak:
Used `LEFT JOIN LATERAL` for combining multiple joins to avoid cross product between tables.

Please review.

Thanks

### DIFF
--- a/src/backend/apidb/writeable_pgsql_selection.cpp
+++ b/src/backend/apidb/writeable_pgsql_selection.cpp
@@ -8,6 +8,7 @@
 #include <sstream>
 #include <list>
 #include <vector>
+#include <boost/lexical_cast.hpp>
 #include <boost/make_shared.hpp>
 #include <boost/ref.hpp>
 #include <boost/shared_ptr.hpp>
@@ -133,6 +134,13 @@ void extract_tags(const pqxx::result::tuple &row, tags_t &tags) {
      tags.push_back(std::make_pair(keys[i], values[i]));
 }
 
+void extract_nodes(const pqxx::result::tuple &row, nodes_t &nodes) {
+  nodes.clear();
+  std::vector<std::string> ids = psql_array_to_vector(row["node_ids"].c_str());
+  for (int i=0; i<ids.size(); i++)
+    nodes.push_back(boost::lexical_cast<osm_nwr_id_t>(ids[i]));
+}
+
 void extract_nodes(const pqxx::result &res, nodes_t &nodes) {
   nodes.clear();
   for (pqxx::result::const_iterator itr = res.begin(); itr != res.end();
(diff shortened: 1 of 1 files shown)


zerebubuth:
Thank you! Your contributions are great, and very welcome :smile:

This isn't a problem with the code - I tested it and it works fine. Unfortunately, OpenStreetMap still [runs PostgreSQL version 9.1 for the main database](https://github.com/openstreetmap/chef/blob/0ff09336c78a7f9dbc446fcb66cb762980fb3ff5/roles/db.rb#L41-L42) which doesn't have the `LATERAL` join type. A couple of solutions I can see to this:

1. We've been wanting to [upgrade to a more recent postgres version](https://github.com/openstreetmap/operations/issues/94#issuecomment-268230456) for a while, and that would fix this. I need to check if there's anything blocking that other than scheduling the downtime.
2. We could figure out if there's another way to write the queries that's backwards-compatible with 9.1.

I think I prefer the first solution, but will have to check how soon that's likely to happen. @pnorman, @tomhughes, what do you think?

tomhughes:
Basically we need to schedule the downtime to do the master flip and version upgrade... It's very much on my to do list though.

zerebubuth:
Great! I've added #132 to make sure Cgimap checks the server version.

@jronak thanks very much for all your help!

pnorman:
> We could figure out if there's another way to write the queries that's backwards-compatible with 9.1.

You could probably do it with `GROUP BY`

jronak:
@pnorman I really missed this out in my mail. I tried using group by on multiple joins but the results with the cross product. Alternative was to use 
`Select .... from current_ways left join ( select way_id, array_agg(k), array_agg(v) from current_way_tags group by way_id) wt on w.id=wt,way_id`
But this would be very inefficient for the smaller selection of ways. 
Kindly let me know if you meant some other way of doing it with group by.
Thanks

pnorman:
```sql
SELECT id, ..., array_agg(k), array_agg(v)
  FROM current_nodes
    LEFT JOIN current_node_tags ON (id=node_id)
  WHERE id = $1
  GROUP BY id
```

I see no particular need to change to this form, just putting it out there as an option

jronak:
Oh, I think I tried this one too anyway I'll it out once again. 
Btw @pnorman is this your work _penorman@mac.com_ ?, Because I sent an email to the same couple of days ago and still haven't heard back. Could please share your work email if it's not the same?

zerebubuth:
I tried the following setup:

```sql
create temporary table foo (id serial primary key);
create temporary table foo_tags (foo_id int references foo(id), tag text);
create temporary table foo_nodes (foo_id int references foo(id), node_id int);

insert into foo (id) values (1);
insert into foo_tags (foo_id, tag) values (1, 'one'), (1, 'two');
insert into foo_nodes (foo_id, node_id) values (1, 1), (1, 2);
```

This should give us a `foo` with "tags" `'one', 'two'` and nodes `1,2`. Running the following query:

```sql
select
  f.id,
  array_agg(t.tag) as tags,
  array_agg(n.node_id) as nodes
from foo f 
  left join foo_tags t on f.id = t.foo_id
  left join foo_nodes n on f.id = n.foo_id
group by f.id;
```

Gives me:

```
 id |       tags        |   nodes   
----+-------------------+-----------
  1 | {two,one,two,one} | {1,1,2,2}
(1 row)
```

I think the `GROUP BY` form is okay for nodes, but for ways and relations where there are joins to both tags and another table, it doesn't work. As @jronak said, we end up with the cross of the tags and third table.

PR #133, opened 2017-02-24
Add check Postgres version during connection setup

### LINKED ISSUE #132: Check PostgreSQL version upon connect
Changes in #131 will require PostgreSQL version 9.3 or later. The previous queries probably relied on an undocumented minimum version, but we should now check and document the version.

We could run [`pqxx::connection_base::server_version`](http://pqxx.org/devprojects/libpqxx/doc/4.0/html/Reference/a00023.html#acdd295f691812773133147530aa6ff21) when we connect to the database and throw an exception if it's less than `90300`? That would provide a more descriptive error message than a query syntax error.

jronak:
Fixes #132 

Thanks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ language: cpp
 compiler: gcc
 cache: ccache
 addons:
-  postgresql: 9.1
+  postgresql: 9.3
   apt:
     packages:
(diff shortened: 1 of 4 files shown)


zerebubuth:
Thanks!

PR #134, opened 2017-03-02
Add default_transaction_read_only 

### LINKED ISSUE #90: Use default_transaction_read_only
Good practices dictate using a user who only has the minimum access required, but running `set default_transaction_read_only = true;` at the start of transaction would be a good idea as an additional security layer.

jronak:
Fixes #90 

Thanks!

### DIFF
--- a/src/backend/apidb/readonly_pgsql_selection.cpp
+++ b/src/backend/apidb/readonly_pgsql_selection.cpp
@@ -547,6 +547,9 @@ readonly_pgsql_selection::factory::factory(const po::variables_map &opts)
   m_connection.set_client_encoding(opts["charset"].as<std::string>());
   m_cache_connection.set_client_encoding(opts["charset"].as<std::string>());
 
+  // set the connection to use readonly transaction.
+  m_connection.set_variable("default_transaction_read_only", "true");
+
   // ignore notice messages
 #if PQXX_VERSION_MAJOR < 4
   m_connection.set_noticer(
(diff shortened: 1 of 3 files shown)


zerebubuth:
I'm not 100% sure, but I think #90 was intended for the writeable backend. We use the readonly one for replica databases, which are read-only already. Having this would help make the tests fail in case we did something non-read-only, so I think we should have it for both backends. @pnorman, what do you think?

pnorman:
I think we should have it on both, and the writable apidb is more important. In production we only run the readonly queries on replicas, but there's no reason it can't be run against a normal database.

jronak:
We have writeable queries in the write apidb, so should the read_only be set only for select queries?

pnorman:
I was thinking that read-only could be used with temporary tables, but was wrong. Trying to do it for only some queries with the writable apidb backend is probably not worth it because it would inhibit connection pooling.

I guess the appropriate way to do this is to run cgimap on a postgres user which only has SELECT access, and then only access to the tables and columns it needs

zerebubuth:
The [PostgreSQL docs](https://www.postgresql.org/docs/current/static/sql-set-transaction.html) say:

> When a transaction is read-only, the following SQL commands are disallowed: `INSERT`, `UPDATE`, `DELETE`, and `COPY FROM` if the table they would write to is not a temporary table; all `CREATE`, `ALTER`, and `DROP` commands; `COMMENT`, `GRANT`, `REVOKE`, `TRUNCATE`; and `EXPLAIN ANALYZE` and `EXECUTE` if the command they would execute is among those listed.

Does that mean that we should set the default transaction to read-only, then start an explicitly writeable transaction to create the temporary tables? All the other creations of `pqxx::work`, which start transactions, should then be read-only, but still able to insert into the temporary table.

pnorman:
Doing setup then going to read-only might be an option. It's just the CREATE statements which are a problem.

jronak:
Updated PR, please review.

zerebubuth:
Wonderful. Thank you!

PR #135, opened 2017-03-04
Fix get_query_string method to handle empty query_string

jronak:
Using lighttpd/1.4.39
The query_string env is empty even when URI has a query, this PR fixes it. 

Thanks!


zerebubuth:
Thanks for spotting this mistake!

PR #136, opened 2017-03-06
Remove per-element queries from read-only APIDB backend

zerebubuth:
Same thing as #131, but for the read-only backend. This separates out the `extract_*` functions into a common source file and shares them between the writeable and read-only backends.

@jronak, it would be very helpful if you could review these changes, please.

@pnorman I'd love to get your opinion on the queries in the read-only backend.


jronak:
LGTM!
This looks very clean and well structured. :-)

zerebubuth:
Thanks @jronak & @pnorman!

> Historic queries to be added in writeable_apidb too ?

I added those before I merged the history branch in #129, so now all calls from read-only and writeable APIDB backends should fetch only one row per element.

PR #137, opened 2017-03-19
Add support for changeset downloads

zerebubuth:
Adds support for the `changeset/#{id}/download` API call.

The element sorting is implemented in the same way as [openstreetmap-website](https://github.com/openstreetmap/openstreetmap-website/blob/834a821d3cc6cbcbc5f267799ede419047740854/app/controllers/changeset_controller.rb#L164-L170), which isn't necessarily exactly the same as the order the elements were committed. But it seems to have sufficed up to this point, so perhaps it's okay :wink:

A small number of tests have been added, and I'd be really interested to know if you think that there are other important areas or corner-cases which should be tested too.

/cc @tomhughes, @jronak & @pnorman - please let me know if you have any advice on how to improve this code, or if you think something's wrong or needs extra testing. Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,7 @@ test/test_apidb_backend_nodes
 test/test_apidb_backend_oauth
 test/test_apidb_backend_historic
 test/test_apidb_backend_changesets
+test/test_apidb_backend_changeset_downloads
 test/test_parse_id_list
 test/test_http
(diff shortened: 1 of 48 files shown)


tomhughes:
I can't say I've looked at every line in detail, but I don't see any obvious issues.

Issue #140, opened 2017-12-26
Add support for diff uploads


zerebubuth:
Now that we have [support for OAuth](https://github.com/zerebubuth/openstreetmap-cgimap/blob/e78578817f4b8f66dfc4e0b91ca1bcfe82cd047c/include/cgimap/oauth.hpp), we can add support for diff (i.e: `osmChange`) uploads. It seems like there's [a considerable improvement to be made by batching queries to the database](https://github.com/openstreetmap/openstreetmap-website/issues/1710#issuecomment-353774921).

Considerations:

* This will be cgimap's first write operation, so we should make sure we've got excellent test coverage and the code has been thoroughly reviewed to avoid corrupting the database.
* Although cgimap supports OAuth for viewing redacted versions of historical elements, it doesn't support HTTP Basic Auth. Only moderators use the redacted-history API, and they can probably live without HTTP Basic Auth. But there might be other people using Basic for the regular user API. Do we want to support that too? Or work with something like [cgimap-ruby](https://github.com/zerebubuth/cgimap-ruby) to allow the auth to be handled at a higher level?

mmd-osm:
Moved previous comments about batching in this issue now:

---

I think it's time for a bit out of the box thinking, and leaving the current implementation behind for a minute. That's why I gave [pyosmium](http://osmcode.org/pyosmium/) and some bulk SQL statements a short try.

The same 8MB diff file was completely parsed in 140ms. Based on this I generated some bulk insert statements, effectively creating all 4300+ nodes in one go:

```sql
BEGIN;
INSERT INTO "current_nodes" ("latitude", "longitude", "changeset_id", "visible", "timestamp", "tile", "version") VALUES
(423712730, -832172277, 19, 't', '2017-12-24 09:15:27.204641', 1234, 1),
(423712670, -834339064, 19, 't', '2017-12-24 09:15:27.204675', 1234, 1),
(423715247, -832172360, 19, 't', '2017-12-24 09:15:27.204684', 1234, 1),
(423715140, -834339084, 19, 't', '2017-12-24 09:15:27.204691', 1234, 1),
(423717479, -834339184, 19, 't', '2017-12-24 09:15:27.204697', 1234, 1),
(423672264, -832171144, 19, 't', '2017-12-24 09:15:27.204704', 1234, 1),
(423672601, -834336964, 19, 't', '2017-12-24 09:15:27.204710', 1234, 1),
[...]

PR #143, opened 2018-05-09
Add support for diff uploads (II)

mmd-osm:
I'm creating this PR as a means to keep both repositories in sync, as outlined in https://github.com/zerebubuth/openstreetmap-cgimap/issues/140#issuecomment-386799449.

It should also facilitate adding comments directly in the source code.


mmd-osm:
I believe now would be a good time to review my changes and merge them back into your branch. As I'm not a collaborator on this repo, I can't raise a "review requested" flag, which I'd like to do...

The implementation should be fairly feature complete by now. Before adding further test cases, feedback on the overall design and implementation would be extremely helpful. The patch is somewhat large, so just take your time.

I already tested {create,modify,delete} - {nodes, ways, relations} is different variations, also with bad data, or non-fulfilled preconditions. Overall guiding principle is to return exactly the same error messages as the Rails port would, as several editors have fancy regex parsers in place.

Thanks!

PR #147, opened 2018-05-21
OAuth parsing failures should be failures, not crashes.

### LINKED ISSUE #146: oauth: boost::spirit::qi::expectation_failure
I was testing some use cases according to the osm website changeset_controller_test.rb file, and noticed an issue with OAuth header parsing. Somehow all test cases use an apostrophe for XML attributes (like in `<node id='#{node.id}'`), and parts of this gets included in the OAuth Auth header. 

Unfortunately, using `'` instead of `"` triggers an HTTP 500 Internal Server error along with an Exception: boost::spirit::qi::expectation_failure. I'm suspecting that this case isn't covered in the parser implementation in oauth.cpp.

 (mod_fastcgi.c.3021) backend died; we'll disable it for 1 seconds and send the request to another backend instead: reconnects: 0 load: 1 

curl statement to reproduce from command line:

```
curl -v 'http://localhost:31337/api/0.6/changeset/876/download' -H 'Accept: */*'  -H 'Content-Type: application/x-www-form-urlencoded' -H $'Authorization: OAuth %3Cdummy%20id="\'-1\'%2F%3E", oauth_consumer_key="U84xxVrHBewaYHehTpaV0Rk3nGhahzRj0zntCe1N", oauth_nonce="Xw1WlI", oauth_signature="32gRmihzmdV47jW2juAL7mIpXkA%3D", oauth_signature_method="HMAC-SHA1", oauth_timestamp="1526814151", oauth_token="FLfg7mbQlV6TAcBSY58YgQz39mpcWgj47J3PZPEx"'
```

BTW: the request was generated by https://github.com/osmlab/osm-auth, maybe there's some issue in that library as well. It shouldn't trigger a crash on cgimap side though.

zerebubuth:
Don't propagate exceptions during OAuth header parsing and validation. Instead, treat it as a validation failure.

Fixes #146.

@mmd-osm, could you try out this branch and let me know if it no longer crashes, please?

### DIFF
--- a/src/oauth.cpp
+++ b/src/oauth.cpp
@@ -460,8 +460,6 @@ boost::optional<std::string> hashed_signature(request &req, secret_store &store)
   return hash;
 }
 
-} // namespace detail
-
 validity::validity is_valid_signature(
   request &req, secret_store &store,
   nonce_store &nonces, token_store &tokens) {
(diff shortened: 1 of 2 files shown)


mmd-osm:
Retest looks good here, payload

```
<dummy id='1'/>
```

now triggers HTTP/400 Bad OAuth request.

I'll also create a follow up issue for osm-auth.

PR #152, opened 2018-07-22
Add support for diff uploads (III)

### LINKED ISSUE #140: Add support for diff uploads

### LINKED ISSUE #162: Missing oauth header file after make install
I'm trying to compile cgimap-ruby (https://github.com/zerebubuth/cgimap-ruby/pull/6) but run into a problem that I think lies in cgimap itself. My script compiles and installs cgimap 0.6.2, and then the patched cgimap-ruby builds against the installed version.

The mkmf log for cgimap-ruby says:

```
have_library: checking for make_apidb_backend() in -lcgimap_apidb... -------------------- no
```

but this is misleading, since the actual complaint is:

```
In file included from /usr/local/include/cgimap/backend/apidb/apidb.hpp:5:0,
                 from conftest.c:3:
/usr/local/include/cgimap/backend.hpp:9:28: fatal error: cgimap/oauth.hpp: No such file or directory
compilation terminated.
```
Indeed that file is not in the /usr/local/include/cgimap/ directory after installation:

```
kitchen@13cffd2d376a:~$  ls /usr/local/include/cgimap/
api06        data_selection.hpp  output_buffer.hpp     request_helpers.hpp
backend      handler.hpp         output_formatter.hpp  routes.hpp
backend.hpp  http.hpp            process_request.hpp   types.hpp
bbox.hpp     logger.hpp          rate_limiter.hpp      zlib.hpp
config.hpp   mime_types.hpp      request.hpp
```
I *think* that's because it's not listed in the [Makefile.am](https://github.com/zerebubuth/openstreetmap-cgimap/blob/eee71075ad27de24ffcdc8b48023e4fe01f86154/Makefile.am#L9-L26) but this is at the limit of my understanding.

mmd-osm:
Follow up for #143, so that all changes appear in the same pull request for easier review.

Fixes #162, closes #140


mmd-osm:
Latest code coverage analysis results: [cgimap_cov.zip](https://github.com/zerebubuth/openstreetmap-cgimap/files/2290190/cgimap_cov.zip)

mmd-osm:
@zerebubuth : I'm planning to merge this PR some time in January 2019, and release it as 0.7.0 version. Hopefully this would leave a bit of time for a high level code review. We could also schedule some session to go through the code on a high level, if that would be helpful in any way.

mmd-osm:
@pnorman & @tomhughes : as you're both familiar with the code base as well, you're of course also invited to add comments, time permitting. In case you have some very specific questions, I'll be glad to provide a high level summary or point you to the relevant parts of the code. This could also include topics that are relevant for  operations, or troubleshooting.

PR #153, opened 2018-07-29
JSON

tomhughes:
This gets the JSON output formatter building again, and enables the JSON output support in the travis builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,9 +19,10 @@ addons:
       - libmemcached-dev
       - libpqxx-dev
       - libxml2-dev
+      - libyajl-dev
 script:
   - ./autogen.sh
-  - ./configure CC="ccache gcc" CXX="ccache g++"
+  - ./configure --enable-yajl CC="ccache gcc" CXX="ccache g++"
   - make
   - make check
 after_failure:
--- a/include/cgimap/json_writer.hpp
+++ b/include/cgimap/json_writer.hpp
@@ -27,9 +27,10 @@ public:
   void end_array();
 
   void entry_bool(bool b);
-  void entry_int(int i);
-  void entry_int(unsigned long int i);
-  void entry_int(unsigned long long int i);
+  void entry_int(int32_t i);
+  void entry_int(int64_t i);
+  void entry_int(uint32_t i);
+  void entry_int(uint64_t i);
   void entry_double(double d);
   void entry_string(const std::string &s);
 
--- a/src/json_writer.cpp
+++ b/src/json_writer.cpp
@@ -83,15 +83,13 @@ void json_writer::end_object() { yajl_gen_map_close(pimpl->gen); }
 
 void json_writer::entry_bool(bool b) { yajl_gen_bool(pimpl->gen, b ? 1 : 0); }
 
-void json_writer::entry_int(int i) { yajl_gen_integer(pimpl->gen, i); }
+void json_writer::entry_int(int32_t i) { yajl_gen_integer(pimpl->gen, i); }
 
-void json_writer::entry_int(unsigned long int i) {
-  yajl_gen_integer(pimpl->gen, i);
-}
+void json_writer::entry_int(int64_t i) { yajl_gen_integer(pimpl->gen, i); }
 
-void json_writer::entry_int(unsigned long long int i) {
-  yajl_gen_integer(pimpl->gen, i);
-}
+void json_writer::entry_int(uint32_t i) { yajl_gen_integer(pimpl->gen, i); }
+
+void json_writer::entry_int(uint64_t i) { yajl_gen_integer(pimpl->gen, i); }
 
 void json_writer::entry_double(double d) {
   // this is the only way, it seems, to use a fixed format for double output.


mmd-osm:
LGTM! The type ambiguity in `  writer->entry_int(elem.changeset);  `  is fixed now.

mmd-osm:
The other question would be, if we should make the format compatible to the Overpass JSON format  http://overpass-api.de/output_formats.html#json - this is already in widespread use, but would require some changes to the current cgimap implementation (which has never been released yet to my knowledge). 

Example: http://overpass-turbo.eu/s/ADD

tomhughes:
Yes the need to decide on what format exactly to return (along with the rails code not being able to do JSON output) is one reason why this has never been enabled.

mmd-osm:
In the meantime I have found out why the gzip compressed response didn't work. There was a missing out->close(), which is getting called in case of xml but not for json.

Fix could be something along those lines:

```
diff --git a/include/cgimap/json_writer.hpp b/include/cgimap/json_writer.hpp
index f1b7d1c..07e7f2a 100644
--- a/include/cgimap/json_writer.hpp
+++ b/include/cgimap/json_writer.hpp
@@ -42,6 +42,7 @@ private:
   // PIMPL idiom
   struct pimpl_;
   pimpl_ *pimpl;
+  boost::shared_ptr<output_buffer> out;
 };
 
 #endif /* JSON_WRITER_HPP */
diff --git a/src/json_writer.cpp b/src/json_writer.cpp
index 9b4b30f..05745a2 100644
--- a/src/json_writer.cpp
+++ b/src/json_writer.cpp
@@ -33,7 +33,7 @@ static void wrap_write(void *context, const char *str, unsigned int len) {
 }
 
 json_writer::json_writer(boost::shared_ptr<output_buffer> &out, bool indent)
-    : pimpl(new pimpl_()) {
+    : pimpl(new pimpl_()), out(out) {
 #ifdef HAVE_YAJL2
   pimpl->gen = yajl_gen_alloc(NULL);
 
@@ -71,6 +71,9 @@ json_writer::~json_writer() throw() {
   yajl_gen_clear(pimpl->gen);
   yajl_gen_free(pimpl->gen);
   delete pimpl;
+
+  if (out != nullptr)
+    out->close();
 }
 
 void json_writer::start_object() { yajl_gen_map_open(pimpl->gen); }

```

mmd-osm:
> The need to decide on what format exactly to return [...] is one reason why this has never been enabled.

We could cc: a number of folks here, or post something on osm dev for some community feedback. And yes, there's sotm18.

tomhughes:
By the way, the reason I started playing with this is https://youtu.be/Jn5VWd7C6_g?t=37m49s which is a lightning talk given at SOTM yesterday.

mmd-osm:
I thought so already, Parsing XML line by line seems like a clever hack, but to me this is really a big no-go. It relies on undocumented line breaks and would force the API to always send those line breaks in the future. If someone ever decides to get rid of those lines breaks (which is totally ok from an XML pov), that new OSM XML parser would immediately break. So yes, consuming JSON right from the start seems like the preferred approach.

I'm a bit sceptic though, if optimizing a 25% portion of the total runtime will have a noticeable impact on perceived performance... but that's another story. We'll find that out soon I'd guess.

Also see my comments here: https://github.com/openstreetmap/iD/pull/5180#issuecomment-408670211, https://github.com/kepta/idly/issues/4

mmd-osm:
I implemented the Overpass API JSON format now, as well as adding the fixes described above to a new pull request: https://github.com/tomhughes/openstreetmap-cgimap/pull/2

Once you merge this pull request into your branch, it should automatically show up in this pull request here. I didn't want to create yet another pull request for the JSON topic in this repo. Let's see if this works out somehow.

mmd-osm:
Let's loop in a few people to get some initial feedback on this topic.

**Question:** Should we consider the Overpass JSON format also for the OSM API JSON output?

**Reasoning:**  Overpass JSON format is around for several years now and supported by a number of tools already. Every overpass turbo user should be familiar with this format. It's also the default that gets shown on the data tab along with the `[out:json]` setting.

Format specification is available here: http://overpass-api.de/output_formats.html#json

In a first step, we're reaching out to developers that are in one way or another involved in OSM editing apps. Please feel free to forward this on other people that might be interested in this discussion as well.

//cc: @zerebubuth , @pnorman , @bhousel, @zverik, @don-vip, @simonpoole, @woodpeck , @drolbr


**Please use the Github voting button if you think it would be a good idea to also implement this format for the OSM API JSON output**

If you think this format is **not** suitable, please consider leaving a comment below.

Thanks for your feedback!

bhousel:
Cool!  Tagging @kepta who gave the talk about the XML parser and @thomas-hervey who just wrote the OSM Notes parser assuming we'd get XML.

On the iD side, we'd love to consume JSON instead of XML, and it would result in a real performance benefit (which @kepta mentioned in his talk). 

@mmd-osm if it would help, I could provide a list of API calls that iD uses.  That would be our priority list.

Overpass-style JSON for the OSM features looks ok to me.  iD would ideally consume JSON for things like Notes and Users too, and even the capabilities call. 😄

simonpoole:
No strong feelings about the choice as the benefits as the subject matter tends to be a) specific to the in browser editors that have traditionally downloaded on panning (yes I know that you can so that in JOSM too) and b) other benefits from using JSON are likely miniscule for non-JS apps. 

So yes I think we should do this because the benefit of silencing some of the commentary from the sidelines and make things somewhat easier for JS devs, but outside of that ....

Zverik:
We wouldn't want to have two JSON formats for OSM data, so I'm strongly in favour of Overpass JSON format. I have used it in my tools, and it's pretty good.

don-vip:
We don't support JSON format in JOSM yet (see [16546](https://josm.openstreetmap.de/ticket/16546)) but I agree with Zverik :)

grischard:
If JSON output format is currently borken and going to change anyway, should it be minified by setting `indent` to false in [choose_formatter.cpp#L246](https://github.com/zerebubuth/openstreetmap-cgimap/blob/master/src/choose_formatter.cpp#L246)?

mmd-osm:
Thanks to everyone for your feedback so far, much appreciated!

@bhousel : 

> @mmd-osm if it would help, I could provide a list of API calls that iD uses. That would be our priority list.
>  iD would ideally consume JSON for things like Notes and Users too, and even the capabilities call.

The focus at the moment is to support all calls returning OSM XML format, as currently implemented by cgimap. This includes:


- the "map" API call,
- single node, way and relation fetches,
- multiple node, way and relation fetches,
- the "full" way and relation calls and
- single node, way and relation history calls,
- single node, way and relation specific version fetches,
- multiple node, way and relation specific version fetches.

I'm not sure if the Apache forwarding rules will include them all right from the start, but I'm pretty sure we will support at least /map once this available.

There's already a first idea how "changset metadata downloads, including discussions" could look like:

/api/0.6/changeset/{{changeset id}}?include_discussion=true

<details>

```json
{
 "version": "0.6",
 "generator": "CGImap 0.6.1 (17758 ubuntu)",
 "copyright": "OpenStreetMap and contributors",
 "attribution": "http://www.openstreetmap.org/copyright",
 "license": "http://opendatacommons.org/licenses/odbl/1-0/",
 "elements": [
  {
   "type": "changeset",
   "id": 1140,
   "created_at": "2018-07-15T19:53:26Z",
   "closed_at": "2018-07-15T20:58:03Z",
   "open": false,
   "user": "mmd2",
   "uid": 1,
   "bounds": {
    "minlat": -33.9133123,
    "minlon": 6.9028606,
    "maxlat": 49.2343776,
    "maxlon": 151.1173123
   },
   "comments_count": 3,
   "comments": [
    {
     "user": "mmd2",
     "uid": 1,
     "created_at": "2018-08-02T19:48:14Z",
     "body": "sdfdsaf"
    },
    {
     "user": "mmd2",
     "uid": 1,
     "created_at": "2018-08-02T19:48:17Z",
     "body": "asdcadcadsfadsf"
    },
    {
     "user": "mmd2",
     "uid": 1,
     "created_at": "2018-08-02T19:48:47Z",
     "body": "<html></html>  {{{ \"\"\""
    }
   ],
   "tags": {
    "comment": "test",
    "created_by": "JOSM/1.5 (14031 SVN de)",
    "source": "test"
   }
  }
 ]
}
```

</details>

--


- changeset downloads is a bit special, as this call returns an osmChange format. This isn't even support in JSON format in Overpass API, hence there isn't an equivalent JSON format definition available.

Other API calls, like notes and capabilities aren't implemented in cgimap but in the Rails port, so those are clearly out of scope atm.

don-vip:
JOSM [14086+](https://josm.openstreetmap.de/changeset/14086/josm) supports Overpass API JSON format.

simonpoole:
@mmd-osm where has that concern been voiced?

mmd-osm:
@simonpoole : it's in the operations issue, https://github.com/openstreetmap/operations/issues/244

don-vip:
This decision is not cool for editor developers like me who spent effort to add JSON support... I'm really disappointed.

mmd-osm:
I really tried my best to convince the Rails camp to run cgimap alongside the Rails port, and be just fine. In the end, it didn't matter to them.

cgimap-ruby is seen by some as the go-to option to integrate cgimap code directly into Rails via some custom Gem. The downside is that the only guru familiar with cgimap-ruby, doesn't have the time to get it ready for production. As sad as it is, we're pretty much stuck here, and I have no idea on how to proceed otherwise.

mmd-osm:
Another attempt of bringing JSON to the Rails port has started in https://github.com/openstreetmap/openstreetmap-website/pull/2221

As I have literally no relevant experience with Rails, we need everyone to help out! You can review code, add comments and suggestions, test it, report errors...

mmd-osm:
Another follow up in https://github.com/openstreetmap/openstreetmap-website/pull/2485

Issue #161, opened 2018-09-13
Boost → C++11 migration 


mmd-osm:
Some ideas for rainy November days: https://github.com/bitcoin/bitcoin/projects/3

Topics covered so far:

- [x] boost::shared_ptr
- [x] boost::scoped_ptr
- [x] boost::noncopyable
- [x] BOOST_FOREACH
- [x] boost::function
- [x] boost::tuple
- [x] boost::bind / ref
- [x] boost date/time functions
- [x] BOOST_ASSERT

mmd-osm:
First part implemented in #164

mmd-osm:
Closing for the time being as most important changes have been implemented already.

PR #164, opened 2019-01-20
Boost → C++11 migration

mmd-osm:
Migrate boost specific features to C++11 standard where possible

Follow up for #161

### DIFF
--- a/include/cgimap/output_writer.hpp
+++ b/include/cgimap/output_writer.hpp
@@ -1,15 +1,29 @@
 #ifndef OUTPUT_WRITER_HPP
 #define OUTPUT_WRITER_HPP
 
+#include <memory>
 #include <string>
 #include <stdexcept>
-#include <boost/noncopyable.hpp>
+
+// due to compatibility issues in gcc 4.8
+template<typename T, typename... Args>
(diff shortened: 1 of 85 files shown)


include/cgimap/output_writer.hpp
```diff
@@ -1,15 +1,29 @@
...
+#include <memory>
 #include <string>
 #include <stdexcept>
-#include <boost/noncopyable.hpp>
+
+// due to compatibility issues in gcc 4.8
+template<typename T, typename... Args>
+std::unique_ptr<T> make_unique(Args&&... args) {
+    return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
+}
```
tomhughes: Do we really need to support a six year old compiler?

mmd-osm: That’s the current default gcc version on Travis CI. I didn’t want to do too many changes at once and bump up gcc in one step.

mmd-osm: I think it would make more sense to change the distribution in .travis.yml to xenial, and use gcc 5.x. Trusty will be eol in about two months anyway.

PR #171, opened 2019-02-04
Refactor: boost date/time functions

### LINKED ISSUE #150: Changeset - incorrect open status with timezone != utc-0
Situation: CGImap runs on a server in UTC+2 timezone. The changeset call /api/0.6/changeset/xxx returns a changeset where the status is "closed", although the changeset will only close 1 hour in the future. For further discussion see details at https://github.com/westnordost/StreetComplete/issues/562#issuecomment-404394683

`now` is initialized with local time (which is UTC+2):

https://github.com/zerebubuth/openstreetmap-cgimap/blob/83185f590317d7f1c5cf9c5ba65b48e6970b5456/src/main.cpp#L183

Comparison doesn't take UTC+2 / UTC+0 difference into account: 

https://github.com/zerebubuth/openstreetmap-cgimap/blob/fdc2523c4293345c0b831740890aa8cd91c0b215/src/output_formatter.cpp#L59-L62

Processing should be based on UTC+0, regardless of the local timezone.

mmd-osm:
This pull request replaces boost date/time functions by their c++11 counterpart. Dependency on boost date-time library is also dropped as part of this pr.

gcc 5.x or newer is required, due to `std::put_time` - requirement fulfilled by Ubuntu Xenial or newer

(Closes #150)


mmd-osm:
@zerebubuth, @tomhughes, @pnorman : I like to drop one boost library dependency in this pull request, and at the same time introduce a minimum gcc version 5.x for cgimap due to `std::put_time`.

I'd appreciate a few additional opinions, if this is a good idea, or if I'd rather should provide some 4.x friendly replacement for `std::put_time` and leave minimum gcc version as is. Thanks!

tomhughes:
No complaints from me.

Zverik:
This would break the RHEL/CentOS 7 support (with gcc 4.8), but I believe nobody runs the Rails Port under that. Ubuntu 16.04 has 5.3, so that change would be okay.

mmd-osm:
Thanks everyone for your feedback, I just merged this pull request.

PR #176, opened 2019-05-04
[Perf] Avoid unnecessary changeset comment extraction, if not requested

mmd-osm:
Prepared statement `extract_changesets` always fetches all related changeset comments from the _changeset_comments_ table, even though it hasn't been requested by the caller. Today, it's cgimap's job to discard that information in case it isn't needed.


mmd-osm:
Reverted in a45717bae2535ee1177b35f40d875f26b0b73e69, because comments_count attribute wasn't filled correctly anymore in case no include_discussion=true parameter was provided.

PR #178, opened 2019-05-25
Use cryptopp path for crypto++ includes

tomhughes:
The cryptopp name is more portable - on Ubuntu it is a link to crypto++ and on Fedora it is the actual name of the directory and crypto++ doesn't work at all.


mmd-osm:
Thanks, good catch! oauth.cpp already uses cryptopp rather than crypto++

PR #179, opened 2019-05-25
Specify columns explictly when inserting into API tables

tomhughes:
If you don't name the columns explicitly then you are depending on the order of the columns in the database which may vary if the history of the database and how it was updated doesn't match the one you are developing against.

### DIFF
--- a/src/backend/apidb/changeset_upload/relation_updater.cpp
+++ b/src/backend/apidb/changeset_upload/relation_updater.cpp
@@ -498,8 +498,8 @@ void ApiDB_Relation_Updater::copy_tmp_create_relations_to_current_relations() {
 
   m.exec(
       R"(
-        INSERT INTO current_relations 
-               (SELECT id, changeset_id, timestamp, visible, version from tmp_create_relations)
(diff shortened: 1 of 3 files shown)


src/backend/apidb/changeset_upload/relation_updater.cpp
```diff
@@ -1353,10 +1353,10 @@ void ApiDB_Relation_Updater::save_current_relations_to_history(
 
   m.prepare("current_relations_to_history",
             R"(   
-                INSERT INTO relations 
-                (SELECT id AS relation_id, changeset_id, timestamp, version, visible
-                 FROM current_relations
-                 WHERE id = ANY($1)) 
+                INSERT INTO relations (relation_id, changeset_id, timestamp, version, visible)
+                SELECT id AS relation_id, changeset_id, timestamp, version, visible
```
mmd-osm: In the current_relations_to_history query below, we use an alias as an `SELECT id AS relation_id`, whereas current_nodes_to_history and current_ways_to_history don't use an alias `SELECT id,`.

It doesn't really make a difference here, it's just something I noticed while looking through the changes.

PR #183, opened 2019-05-26
Fix history api returns wrong/duplicate tag data for ways

### LINKED ISSUE #181: history api returns wrong tag data for ways
The api call
​https://www.openstreetmap.org/api/0.6/way/158601617/history
returns a lot of duplicated tags for each version, it seems that a join is done without using the version number. Attached is the result.
[history.txt](https://github.com/zerebubuth/openstreetmap-cgimap/files/3220070/history.txt)

mmd-osm:
Fixes #181


tomhughes:
I guess we should make sure we have a test for this case as well? I assume we are able to test readonly vs read-write mode separately?

mmd-osm:
I also thought about adding a unit test just a minute ago, will do in a separate commit.

Issue #189, opened 2019-05-30
Changeset upload with basic auth results in "401 Unauthorized"


woodpeck:
When changeset upload on the OSM API was switched to CGImap, my Perl scripts that upload changesets with HTTP_BASIC_AUTH have stopped working. For example, after successfully having created changeset 70762083, I tried to upload

```
<osmChange version='0.6'>
<delete>
 <node id="6285571178" visible="true" version="2" changeset="70762083" timestamp="2019-03-14T21:24:38Z" user="mueschel" uid="616774" lat="-0.7734650" lon="30.9558450"/>
</delete>
</osmChange>
```
with
```
perl changeset.pl upload 70762083 < changeset.xml
```
and got
```
POST https://api.openstreetmap.org/api/0.6/changeset/70762083/upload... 401 Unauthorized (22b)
Response:
User is not authorized
cannot upload changeset: 401 Unauthorized
User is not authorized
```
I tried a few different values for the "changeset", "uid" and "user" attributes (trying both the values of the node to be deleted and the changeset in which the deletion was supposed to happen) but to no avail; 401 Unauthorized throughout.

tomhughes:
Just to be clear basic auth appears to be working for other requests (at least I got a 401 if I use the wrong password and a 200 if I use the right one) but for upload it does not appear to be working at all and everybody is getting a 401 response I think.

I have confirmed that the HTTP_AUTHORIZATION header is present in the cgimap FCGI request.

I have moved upload back to rails on the production servers for now.

mmd-osm:
Basic Auth works for me in JOSM without any issue. We need to further investigate what the issue is with your Perl script.

tomhughes:
It's not just his script. I've seen it with JOSM uploads as well.

mmd-osm:
I believe we would have seen many more reports if this is a general issue with Basic Auth (assuming it is still used by a fair amount of users).

All of the basic auth logic is in this class https://github.com/zerebubuth/openstreetmap-cgimap/blob/master/src/basicauth.cpp - I'm wondering what the best way would be to reproduce the issue, as I obviously don't have access to either the user/passwords and the salted hashes on the db tables.

tomhughes:
So I don't of course know that they aren't genuine wrong passwords but I have 70 in about three hours this morning on one of three servers from at least:

* JOSM
* Merkaartor
* OsmAnd
* archINFORM-Wikidata-Enricher
* osmtools

[...]

PR #190, opened 2019-05-30
Basic auth should also consider user email and lower case variants

mmd-osm:
Keeps authentication is line with user.rb, authenticate method
See #189


mmd-osm:
Unit tests are also in and green -> merging into master now.

Issue #84, opened 2013-11-01
Make MAX_NODES and MAX_AREA values configurable


pnorman:
Currently we define

``` c++
#define MAX_AREA 0.25
#define MAX_NODES 50000
```

with the logic

``` c++
int num_nodes = sel.select_nodes_from_bbox(b, MAX_NODES);
if (num_nodes > MAX_NODES) {
  throw http::bad_request(
    (format("You requested too many nodes (limit is %1%). "
            "Either request a smaller area, or use planet.osm")
            % MAX_NODES).str());
}
```

We've said it would be good to the max values configurable before, but now **I** have a project that could use it, so I have some motivation to fixing it.

I think it makes sense to change where the logic is being evaluated at the same time.

You can imagine a backend which caches in a tiled schema so it makes sense to return a larger area than requested for small requests. This could mean that for an area with an exceptionally high node density that the number of nodes selected will always be >MAX_NODES.

Another example is a backend that somehow knows that a particular request will be more than MAX_NODES without running it. It would be foolish to run it, but the number of nodes selected for select_nodes_from_bbox would be 0.

This could be accommodated by changing `select_nodes_from_bbox(b,MAX_NODES)` to throw `http::bad_request` itself

PR #222, opened 2020-03-12
Make MAX_NODES and MAX_AREA values configurable

bmarchant:
This PR creates a configuration options class that parses a configuration XML file so that code like this from `include/cgimap/util.hpp` doesn't exist hard-coded but can be configured if desired:
```
const int CHANGESET_MAX_ELEMENTS = 10000;
const int WAY_MAX_NODES = 2000;
const long SCALE = 10000000;
```
The `Options` class is a singleton that doesn't need to be passed all throughout the code to access the configuration options.  The options are parsed once and used throughout.

The new configuration file `config/cgimap.xml.orig` is copied to `/usr/local/etc/cgimap/` (or `$(sysconfdir)` in automake) during the `make install` process.  This does not overwrite the current configuration file because of the `.orig` extension.  It is not required nor read unless it is renamed to `cgimap.xml`.

The new `options.hpp` and `options.cpp` files were run through `clang-format` to ensure correct code formatting.

More notes follow within the code itself.

References [Issue 84](https://github.com/zerebubuth/openstreetmap-cgimap/issues/84)


config/cgimap.xml.orig
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<cgimap copyright="OpenStreetMap and contributors" attribution="http://www.openstreetmap.org/copyright" license="http://opendatacommons.org/licenses/odbl/1-0/">
+  <settings>
+    <map_nodes maximum="50000"/>
+    <area maximum="0.25"/>
+    <changeset maximum_elements="10000"/>
+    <waynodes maximum="2000"/>
+    <relation_members maximum="32000"/>
+    <element_tags maximum="5000"/>
```
bmarchant: These two options were added so that the current PR "Enforce limits for tags and relation members #174" is already integrated.
https://github.com/zerebubuth/openstreetmap-cgimap/pull/174

include/cgimap/api06/changeset_upload/way.hpp
```diff
@@ -24,15 +25,15 @@ class Way : public OSMObject {
...
 
     try {
-	_waynode = std::stol(waynode);
+      _waynode = std::stol(waynode);
     } catch (std::invalid_argument& e) {
-	throw xml_error("Way node is not numeric");
+      throw xml_error("Way node is not numeric");
     } catch (std::out_of_range& e) {
-	throw xml_error("Way node value is too large");
+      throw xml_error("Way node value is too large");
```
bmarchant: In a few of the files there were a mixture of tabs and spaces that were fixed during formatting.

include/cgimap/api06/changeset_upload/way.hpp
```diff
@@ -55,12 +56,12 @@ class Way : public OSMObject {
                 .str());
       }
 
-      if (m_way_nodes.size() > WAY_MAX_NODES) {
+      if (m_way_nodes.size() > Options::get_instance().get_way_max_nodes()) {
```
bmarchant: This is the basis of the PR, replace all constant expressions with a call to the `Options` object to get the correct configuration value.

include/cgimap/options.hpp
```diff
@@ -0,0 +1,74 @@
...
+  int map_max_nodes = 50000;
+  double map_area_max = 0.25;
+  int changeset_max_elements = 10000;
+  int way_max_nodes = 2000;
+  int relation_max_members = 32000;
+  int element_max_tags = 5000;
+  long bbox_scale = 10000000L;
+  std::string timeout_open_max = "1 day";
+  std::string timeout_idle = "1 hour";
+  long input_buffer_max = 50000000L;
```
bmarchant: Default values are given here in case no configuration file is present.

openstreetmap-cgimap.1
```diff
@@ -55,6 +56,10 @@ Write pid to \fIPIDFILE\fR.
 .BR \-\-logfile =\fILOGFILE\fR
 File to write log messages to.
 .TP
+.BR \-\-configfile =\fICONFIGFILE\fR
+File to read configuration values from.
+Default is '/usr/local/etc/cgimap/cgimap.xml'.
+.TP
```
bmarchant: Added `--configfile` command line argument, description, and default value to `man` page.

src/Makefile.am
```diff
@@ -129,15 +129,16 @@ libcgimap_fcgi_la_LIBADD=@FCGI_LIBS@
...
 	backend.cpp \
+	basicauth.cpp \
 	bbox.cpp \
 	choose_formatter.cpp \
 	data_selection.cpp \
 	handler.cpp \
 	http.cpp \
 	logger.cpp \
 	mime_types.cpp \
-        basicauth.cpp \
```
bmarchant: Alphabetized and formatted source listings.

src/main.cpp
```diff
@@ -102,6 +103,7 @@ static void get_options(int argc, char **argv, po::variables_map &options) {
     ("instances", po::value<int>()->default_value(5), "number of daemon instances to run")
     ("pidfile", po::value<string>(), "file to write pid to")
     ("logfile", po::value<string>(), "file to write log messages to")
+    ("configfile", po::value<string>()->default_value("/usr/local/etc/cgimap/cgimap.xml"), "configuration file")
```
bmarchant: Added command line argument and default value.

tomhughes:
What's the logic behind doing this via a file when all the rest of the configuration is via switches and/or environment variables?

Either we should move everything to a file or this should be configured the same way as existing options but have two different ways of configuring things doesn't seem very sensible.

bmarchant:
Here are my thoughts on configuration file vs. command line arguments and why I chose to implement these changes in a config file.

A configuration file gives a central place that is associated with the actual executable (in this case a file called `/usr/local/etc/cgimap/cgimap.xml`) to configure the executable instead of in other locations depending on the installation when passing configuration options via the command line. 
 Adding 9 more potential command line arguments to the command line doesn't seem like the right approach to me.

My recent experience in setting up a `cgimap` server for a project that I'm working on started with testing from the command line, then moving to a VM running `lighttpd` for the proof of concept phase, and finally converting to using it in Apache to replace the rails port calls.  This required a lot of reconfiguring to get each one working.  My logic goes as follows since I just went through this process.

For testing purposes I figure most developers run `cgimap` from the command line and have to create a shell script with all of the parameters set and run that.  A file like that is not in GitHub so all developers must create their own or retype (or use history) each time they run it.

Then I moved to the second option for running a quick and dirty `cgimap` server is using `lighttpd` and the configuration for that setup is done in the `/etc/lighttpd/lighttpd.conf` file part of which is below where the environment variables are set in the `bin-environment` section of `fastcgi.server` below from the sample file in GitHub:
```
fastcgi.server = ( ".map" =>
  (( "socket" => "/tmp/map-fastcgi.socket",
     "bin-path" => "/home/matt/Programming/openstreetmap/cgimap/openstreetmap-cgimap",
     "docroot" => "/",
     "min-procs" => 1,
     "max-procs" => 1,
     "check-local" => "disable",
     "bin-environment" => (
#       "CGIMAP_HOST" => "localhost",
       "CGIMAP_DBNAME" => "apidb",
       "CGIMAP_LOGFILE" => "/home/matt/Programming/openstreetmap/cgimap/log",
#       "CGIMAP_USERNAME" => "matt",
#       "CGIMAP_PASSWORD" => ""
     )
  ))
)
```
Finally I began to integrate this with Apache and Phusion on an Openstreetmap rails setup and the configuration is in a totally different configuration file, not in Apache but in the init script (sample at `scripts/cgimap.init`) to start the daemon as a service.
```
CGIMAP_HOST=localhost; export CGIMAP_HOST
CGIMAP_DBNAME=openstreetmap; export CGIMAP_DBNAME
CGIMAP_USERNAME=openstreetmap; export CGIMAP_USERNAME
CGIMAP_PASSWORD=openstreetmap; export CGIMAP_PASSWORD
CGIMAP_PIDFILE=cgimap.pid; export CGIMAP_PIDFILE
CGIMAP_LOGFILE=cgimap.log; export CGIMAP_LOGFILE
CGIMAP_MEMCACHE=localhost; export CGIMAP_MEMCACHE
CGIMAP_RATELIMIT=102400; export CGIMAP_RATELIMIT
CGIMAP_MAXDEBT=250; export CGIMAP_MAXDEBT
start() {
  start-stop-daemon --start --chuid rails --exec /home/rails/bin/openstreetmap-cgimap -- --daemon --port=8000 --instances=30
}
```
I found that I couldn't use this file because with the latest release of Ubunutu `init d` is no longer used but replaced by `systemd` which requires a `cgimap.service` file, an exercise left for the user to create.  The version of Ubuntu now in use with Vagrant is `bionic` 18.04 as of [30 DEC 2019](https://github.com/zerebubuth/openstreetmap-cgimap/commit/34b2cd081ccc757aab60211a07d59f5c32612311#diff-23b6f443c01ea2efcb4f36eedfea9089). 

I do think that moving most configuration settings to a file is the right thing to do, others may disagree, and this is a start.  Since `cgimap` is being used extensively, in production with OpenstreetMap, removing all support for environment variables and command line arguments is not tenable at this moment.  I suggest adding support for removing these a few at a time or migrating them all to the config file while allowing the command line to override the config file values.

This code is my solution to tuning my own `cgimap` instance and finding that some of the values give better performance for my particular use-cases.  I can simply update the configuration file and bounce the server and run my tuning tests.

tomhughes:
Oh I'm not disputing that a configuration file is probably a good idea, just that we should move to it wholesale rather than splitting configuration over multiple locations.

bmarchant:
We could do it all at once or little by little, it is up to the powers that be.  I'm more than willing to extend this further if that is the agreed upon direction.

mmd-osm:
Thanks a lot for working on this topic. I've added a few comments based on a rather brief review, I still need to do a more thorough review and try this code out!

I agree we should look into offering an option to specify all configuration settings in one file, rather than having parts of the config as environment variables/command line arguments only.

I'm a bit undecided what to do when both environment variables/command line args and configuration files want to set the same parameter: some of the options I can think of could be  raising some warning/error, or using environment/command line args with higher prio, and only fall back to the configuration file if no parameter value is available (and a config file has been given on the command line).

bmarchant:
Part 2, Tom makes a good point.  Here are the rest of the settings in the configuration file.

@mmd-osm I've removed the limits on relation members and tag max.  And as for the higher priority, I think that command line arguments should have the priority.  Makes sense that one should see the setting change if a command line argument is added instead of being overridden by the config file.  So that is the way I've updated the code, first read the config file and then overwrite any values found on the command line.  I'll add some other comments to the code.

config/cgimap.service
```diff
@@ -0,0 +1,13 @@
...
+After=network.target
+
+[Service]
+Type=forking
+ExecStart=/usr/local/bin/openstreetmap-cgimap --daemon
+PIDFile=/var/run/cgimap.pid
+
+[Install]
+WantedBy=multi-user.target
+Alias=cgimap
```
bmarchant: Ubuntu 18.04 (latest used by Vagrant) now uses `systemd` instead of `init` so I created a basic service file for that just like `cgimap.conf` for `init`.

include/cgimap/backend.hpp
```diff
@@ -24,13 +24,13 @@ struct backend {
   options() const = 0;
   // create a data selection factory from the arguments passed to cgimap.
   virtual std::shared_ptr<data_selection::factory>
-  create(const boost::program_options::variables_map &) = 0;
+  create() = 0;
```
bmarchant: No longer necessary to pass `variables_map` object all over the code base since the `Options` class is a singleton and is accessible everywhere it is needed.

src/main.cpp
```diff
@@ -258,34 +285,20 @@ int main(int argc, char **argv) {
...
+      if ((socket = fcgi_request::open_socket(config_options.get_socket_path(), 5)) < 0) {
         throw runtime_error("Couldn't open FCGX socket.");
       }
-      // fall back to the old --port option if socket isn't available.
-    } else if (options.count("port")) {
-      std::ostringstream sock_str;
-      sock_str << ":" << options["port"].as<int>();
-      if ((socket = fcgi_request::open_socket(sock_str.str(), 5)) < 0) {
-        throw runtime_error("Couldn't open FCGX socket (from port).");
-      }
```
bmarchant: No need to fall back on the old port, this is all done in the `Options` class.

mmd-osm:
Thank you for your continued patience... 

Extending a bit on the idea to split up the Options class to have a dedicated XML parsing class, I could imagine something like https://godbolt.org/z/mP7J9C with an abstract base class used as callback interface.

XML options parsing is quite complicated and a crucial part of the whole application. I think it would make sense to also have dedicated unit tests just covering that part.

For unit tests, we would define some mock callback class which would validate all parameter value that have been found during parsing. It should cover positive and in particular error test cases that you're already handling in the parser today. Input for unit tests should probably come from strings, rather than files in that case.

mmd-osm:
In the meantime, I spent some more time on code review and came to the conclusion that we need to disentangle the Options class a bit. Currently, the Options class includes:
* a single pattern to manage an instance of Option, that includes a number of parameter values
* exposes a get_instance method for consumers to access the config settings
* methods and state handling for an XML parser
* an XML parser, which includes fallback value handling
* default values for all parameters
* parameter value validations
* a method to derive a Postgresql connection string from a number of parameter values
* override logic to blend default, command line, and config file values into a final value
* a method return a summary statement of all configuration options to print on the screen

Also, the existing mechanism to pass parameter values around was moved over to the Options class.

I think, we have to go back to the drawing board and think about some ways to break things up, and make them testable.

In a first iteration, I decided to drop all methods in the global options class, except for those 8 parameter values, which have been hard-coded before. Also, I dropped the singleton "get_instance" pattern, an implementation detail which I don't want to expose to outside consumers.

The options class itself now consists of a few static methods, which delegate parameter value retrieval to another class (based on an abstract base class). When instantiating the options class, a class with default values is automatically used for initialization. It can be replaced by another class later on (= a new implementation can be injected into the options class).

I'm very reluctant to move all the database settings away from "options" as is today, and would rather  add any values we find the config file to  "options". I really want to keep the footprint of those global structures as small as possible.

So here's what the current, stripped down approach looks like: https://github.com/mmd-osm/openstreetmap-cgimap/commit/3ac32b037013846a291068ae37daebdc27906aa6

I think in a next step, I would look into reusing your work on the XML parser, turn it into a standalone class and add some unit tests.

mmd-osm:
New implementation moved to #232 now. Needs more testing and documentation.

config/cgimap.service
mmd-osm: I think this would be a topic for another pull request. I will skip this for now.

README
```diff
@@ -103,36 +103,64 @@ CGImap with lighttpd you will need to install lighttpd:
...
 
 An example of this can be found in
 [OSM Chef](http://git.openstreetmap.org/chef.git/blob/HEAD:/cookbooks/web/recipes/cgimap.rb).
 
+For more recent operating systems using systemd instead of init
+`config/cgimap.service` is provided to run CGImap as a daemon.  To
+use it modify the paths, set environment variables or the settings
+in the config file, copy it to `/etc/systemd/system/cgimap.service`.
+Change the mode to 755, and owner:group to root:root.  To enable the service us
+`systemctl enable cgimap` and to start the service `systemctl start cgimap`.
```
mmd-osm, line 127: This section seems like a topic for another pull request.

README
```diff
@@ -103,36 +103,64 @@ CGImap with lighttpd you will need to install lighttpd:
...
 to the executable. See `./openstreetmap-cgimap --help` for a list of options.
 To convert a command line option to an environment variable append `CGIMAP_` to
 the option and capatalize it. For example, the option `--dbname` becomes the
 environment variable `CGIMAP_DBNAME`.
 
+These connection parameters (and other settings) can be passed to the executable via
+command line arguements or in the XML configuration file located at `config/cgimap.xml.orig`.
+The location of this file can be set with the `--configfile` option or `make install`
+installs the file to `/usr/local/etc/cgimap/cgimap.xml.orig` and can be updated and renamed
+to `/usr/local/etc/cgimap/cgimap.xml` which is the default location that the cgimap
```
mmd-osm, line 139: I will skip this part of installing an XML file.

config/cgimap.xml.orig
```diff
@@ -0,0 +1,41 @@
...
+    </map>
+    <changeset>
+      <timeout open="1 day" idle="1 hour"/>
+    </changeset>
+    <changeset_upload>
+      <changeset maximum_elements="10000"/>
+      <waynodes maximum="2000"/>
+    </changeset_upload>
+  </settings>
+</cgimap>
```
mmd-osm, line 41: XML will need to be replaced by ini format file

include/cgimap/backend.hpp
mmd-osm: I left those as is

src/options.cpp
```diff
@@ -0,0 +1,408 @@
...
+  if (!get_backend_username(type).empty())
+    ostr << " user=" << get_backend_username(type);
+  if (!get_backend_password(type).empty())
+    ostr << " password=" << get_backend_password(type);
+  if (!get_backend_port(type).empty())
+    ostr << " port=" << get_backend_port(type);
+  return ostr.str();
+}
+
+void Options::on_start_element(const char *element, const char **attrs) {
```
mmd-osm, line 126: I left out all XML processing and let boost program options do the work instead. This not only saves us some extra parsing logic, but also makes sure that command line, environment and configuration file use the same parameter definition without a need to keep multiple places in sync.

src/options.cpp
```diff
@@ -0,0 +1,408 @@
...
+      } else if (!std::strcmp(element, "socket")) {
+        socket_path = check_attributes(attrs, "path", socket_path);
+      } else if (!std::strcmp(element, "memcache")) {
+        memcache_spec = check_attributes(attrs, "spec", memcache_spec);
+      } else if (!std::strcmp(element, "ratelimit")) {
+        rate_limit = check_attributes(attrs, "rate", rate_limit);
+        if (rate_limit <= 0)
+          throw std::invalid_argument("Invalid rate limit value");
+        max_debt = check_attributes(attrs, "maxdebt", max_debt);
+        if (max_debt <= 0)
```
mmd-osm, line 162: Parameter value validation is good. I used the same kind of checks.

src/options.cpp
```diff
@@ -0,0 +1,408 @@
...
+  return info;
+}
+
+std::string Options::get_configuration_options() const
+{
+  std::ostringstream ostr;
+  ostr << std::left << std::boolalpha;
+  std::_Setw w = std::setw(25);
+  /** General Settings */
+  ostr << "--- General Settings ---" << std::endl
```
mmd-osm, line 343: I left out this method as I currently don’t see a need for it. I think maybe there’s some easy way to let boost program options do something similar.

PR #232, opened 2020-06-18
Make MAX_NODES and MAX_AREA values configurable (2)

### LINKED ISSUE #84: Make MAX_NODES and MAX_AREA values configurable

mmd-osm:
Revised implementation for #222 

Closes #222, closes #84.

### DIFF
--- a/include/cgimap/api06/changeset_upload/way.hpp
+++ b/include/cgimap/api06/changeset_upload/way.hpp
@@ -2,6 +2,7 @@
 #define WAY_HPP
 
 #include "osmobject.hpp"
+#include "cgimap/options.hpp"
 
 #include <string>
 #include <vector>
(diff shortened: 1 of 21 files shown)


mmd-osm:
Thanks again to @bmarchant for the original work on #222, and apologies for the long time it took to review all the code. I ended up with a solution with less changes to the existing code, and as I tend to be a bit lazy at times, I opted to delegate as much work as possible to the existing boost program options library we already have in place.

Issue #199, opened 2019-07-12
CGIMAP_* environment parameter blocks --help


mmd-osm:
Setting some CGIMAP_* environment variable seems to impact boost program options to no longer accept  `--help` as a command line option.

```
export CGIMAP_FOO=1
./openstreetmap-cgimap --help
Exception: unrecognised option
```

PR #237, opened 2020-10-01
CGIMAP_* environment parameter blocks --help

bmarchant:
References https://github.com/zerebubuth/openstreetmap-cgimap/issues/199

Refactored `environment_option_name` into a validation lambda that checks against the options in `po::options_description`.

Fixed up a few places in the code and Makefile where tabs and spaces didn't match and was causing visual alignment issues.

Updated `man` page to include config file description and oauth/update database command line arguments.

### DIFF
--- a/include/cgimap/http.hpp
+++ b/include/cgimap/http.hpp
@@ -15,10 +15,6 @@
 #endif
 #include "cgimap/output_buffer.hpp"
 
-// Maximum bytes
-const unsigned long STDIN_MAX = 50000000;    // TODO: configurable parameter?
-
-
 /**
  * Contains the generic HTTP methods and classes involved in the
  * application. CGI-specific stuff is elsewhere, but this stuff
--- a/openstreetmap-cgimap.1
+++ b/openstreetmap-cgimap.1
@@ -9,6 +9,7 @@ openstreetmap-cgimap \- FCGI version of the OpenStreetMap API
 [\fB\-\-instances \fIINSTANCES\fR]
 [\fB\-\-pidfile \fIPIDFILE\fR]
 [\fB\-\-logfile \fILOGFILE\fR]
+[\fB\-\-configfile \fICONFIGFILE\fR]
 [\fB\-\-memcache \fISPEC\fR]
 [\fB\-\-ratelimit \fILIMIT\fR]
 [\fB\-\-maxdebt \fIDEBT\fR]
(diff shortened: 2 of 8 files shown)


include/cgimap/http.hpp
```diff
@@ -15,10 +15,6 @@
 #endif
 #include "cgimap/output_buffer.hpp"
 
-// Maximum bytes
-const unsigned long STDIN_MAX = 50000000;    // TODO: configurable parameter?
-
-
```
bmarchant, line 21: No longer used, it is replaced by `global_settings::get_payload_max_size()`

openstreetmap-cgimap.1
```diff
@@ -9,6 +9,7 @@ openstreetmap-cgimap \- FCGI version of the OpenStreetMap API
 [\fB\-\-instances \fIINSTANCES\fR]
 [\fB\-\-pidfile \fIPIDFILE\fR]
 [\fB\-\-logfile \fILOGFILE\fR]
+[\fB\-\-configfile \fICONFIGFILE\fR]
```
bmarchant, line 12: `--configfile` missing from list in man page

openstreetmap-cgimap.1
```diff
@@ -25,12 +26,19 @@ openstreetmap-cgimap \- FCGI version of the OpenStreetMap API
...
+[\fB\-\-oauth\-password \fIOAUTHPASS\fR]
+[\fB\-\-oauth\-charset \fIOAUTHCHARSET\fR]
+[\fB\-\-oauth\-dbport \fIOAUTHPORT\fR]
+] [
+[\fB\-\-update\-dbname \fIUPDATEDBNAME\fR]
+[\fB\-\-update\-host \fIUPDATEHOST\fR]
+[\fB\-\-update\-username \fIUPDATEUSER\fR]
+[\fB\-\-update\-password \fIUPDATEPASS\fR]
+[\fB\-\-update\-charset \fIUPDATECHARSET\fR]
+[\fB\-\-update\-dbport \fIUPDATEPORT\fR]
```
bmarchant, line 41: oauth/update parameters omitted.

src/main.cpp
```diff
@@ -74,23 +74,6 @@ static string get_generator_string() {
...
-    std::transform(name.begin() + 7, name.end(),
-                   std::back_inserter(option),
-                   [](unsigned char c) {
-                     return c == '_' ? '-' : std::tolower(c);
-                   });
-  }
-
-  return option;
-}
-
```
bmarchant, line 93: Replaced with lambda expression below that also validates against known option names.

src/main.cpp
```diff
@@ -136,7 +119,24 @@ static void get_options(int argc, char **argv, po::variables_map &options) {
...
+  po::store(po::parse_environment(desc,
+    [desc](const std::string &name) {
+          std::string option;
+          // convert an environment variable name to an option name
+          if (name.substr(0, 7) == "CGIMAP_") {
+            std::transform(name.begin() + 7, name.end(),
+                           std::back_inserter(option),
+                           [](unsigned char c) {
+                           return c == '_' ? '-' : std::tolower(c);
+                   });
```
bmarchant, line 131: Identical to `environment_option_name` function removed above.

src/main.cpp
```diff
@@ -136,7 +119,24 @@ static void get_options(int argc, char **argv, po::variables_map &options) {
...
+            for (const auto &d : desc.options()) {
+              if (d->long_name() == option)
+                return option;
+            }
+            std::cout << "Ignoring unknown environment variable: " << name << std::endl;
```
bmarchant, line 136: Validating against all option names from the `options_description` object `desc`.

mmd-osm:
Thanks! I wonder if it would make sense to move the "--help" logic right after the command line parsing. Otherwise you would see an "Ignoring unknown environment variable" message even in case of "--help". I also don't expect anyone to set CGIMAP_HELP to get those usage instructions.

```
  po::store(po::parse_command_line(argc, argv, desc), options);

  if (options.count("help")) {
    std::cout << desc << std::endl;
    output_backend_options(std::cout);
    exit(1);
  }  
```

src/main.cpp
```diff
@@ -136,7 +119,24 @@ static void get_options(int argc, char **argv, po::variables_map &options) {
   desc.add(expert);
 
   po::store(po::parse_command_line(argc, argv, desc), options);
-  po::store(po::parse_environment(desc, environment_option_name), options);
+  po::store(po::parse_environment(desc,
+    [desc](const std::string &name) {
```
mmd-osm, line 123: Maybe capture desc by reference instead?
```
  po::store(po::parse_environment(desc,
    [&desc](const std::string &name) {
```

bmarchant: Good point, no need to pass by value there.

bmarchant:
Setting `CGIMAP_HELP` doesn't make sense so the help function call can precede the environment variable checks.

PR #246, opened 2021-05-15
Adding OAuth2 support

### LINKED ISSUE #238: Adding OAuth2 support
See https://github.com/openstreetmap/openstreetmap-website/issues/1408

mmd-osm:
Closes #238

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -19,6 +19,7 @@ cgimap_include_HEADERS = \
 	include/cgimap/mime_types.hpp \
 	include/cgimap/basicauth.hpp \
 	include/cgimap/oauth.hpp \
+	include/cgimap/oauth2.hpp \
 	include/cgimap/output_buffer.hpp \
 	include/cgimap/output_formatter.hpp \
 	include/cgimap/process_request.hpp \
@@ -108,7 +109,7 @@ endif
 
 TESTS = test/map.testcore test/node.testcore test/anon.testcore test/way.testcore test/relation.testcore
 TESTS += test/empty.testcore test/way_full.testcore test/relation_full.testcore
-TESTS += test/test_parse_id_list test/test_basicauth test/test_oauth test/test_http test/test_parse_time
+TESTS += test/test_parse_id_list test/test_basicauth test/test_oauth test/test_oauth2 test/test_http test/test_parse_time
 TESTS += test/changesets.testcore test/message.testcore test/routes.testcore
 TESTS += test/redactions.testcore
 TESTS += test/test_parse_osmchange_input
@@ -124,6 +125,7 @@ if ENABLE_APIDB
 TESTS += \
 	test/test_apidb_backend_nodes \
 	test/test_apidb_backend_oauth \
+	test/test_apidb_backend_oauth2 \
 	test/test_apidb_backend_historic \
 	test/test_apidb_backend_changesets \
 	test/test_apidb_backend_changeset_downloads \
--- a/include/cgimap/backend/apidb/oauth_store.hpp
+++ b/include/cgimap/backend/apidb/oauth_store.hpp
@@ -19,6 +19,7 @@ public:
   bool allow_write_api(const std::string &token_id);
   boost::optional<osm_user_id_t> get_user_id_for_token(const std::string &token_id);
   std::set<osm_user_role_t> get_roles_for_user(osm_user_id_t id);
+  boost::optional<osm_user_id_t> get_user_id_for_oauth2_token(const std::string &token_id, bool& expired, bool& revoked, bool& allow_api_write);
 
 private:
   pqxx::connection m_connection;
--- a/include/cgimap/oauth.hpp
+++ b/include/cgimap/oauth.hpp
@@ -17,6 +17,7 @@ namespace oauth {
 struct secret_store {
   virtual boost::optional<std::string> consumer_secret(const std::string &consumer_key) = 0;
   virtual boost::optional<std::string> token_secret(const std::string &token_id) = 0;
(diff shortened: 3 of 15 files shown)


tomhughes:
What are the plans as far as merging this and doing a release are concerned?

mmd-osm:
I wanted to give other developers a chance to try this out on oauth2.apis.dev.openstreetmap.org before doing a release. Somehow there isn't as much activity there as I was hoping to see.

tomhughes:
Well the problem is that there are approximately zero OAuth 2 supporting clients at the moment so who exactly is going to test it?

Either we have to start a major outreach campaign to get people to starting porting clients to OAuth 2 using the test instance or we have to just go ahead and fix the bugs later - as there will be no users for the live API in the first instance I would have thought the second could work, that as people write clients and test the we can fix any issues?

simonpoole:
Make an basic auth end of support announcement for openstreetmap.org? At the same time suggesting that the effected apps use OAuth2.

tomhughes:
@simonpoole that's really a discussion to have on the web site tracker not here but I'd rather get OAuth 2 out there and get some experience with it first before worrying about that.

simonpoole:
@tomhughes it was just a suggestion to make the "major outreach campaign" easy :-) (yes I do have a secret agenda in this context that belongs on the website repo).

mmd-osm:
@Firefishy sent out an announcement on Slack OSM US, dev channel back in April I believe. At least JOSM already has an open issue for it: https://josm.openstreetmap.de/ticket/20768 . I haven't seen much activity wrt OAuth2 for other editing apps.

What would be the planned timeline to merge OAuth2 into the Rails port?

tomhughes:
Well I was about to do it on Saturday when I realised we needed to do this first, or more precisely I thought you had already merged this and came here to ask about getting a release and build and found I was wrong.

mmd-osm:
I think it would be good to double check this regular expression for Bearer tokens: https://github.com/zerebubuth/openstreetmap-cgimap/pull/246/files#diff-0b612f1272c9dabd3419184a64a8cba1b5901b890b2e8efdda253d107951d466R38

According to my understanding, doorkeeper uses SecureRandom.urlsafe_base64 to generate Bearer tokens. As the padding parameter is left unspecified, no trailing "=" characters will be added. Is this understanding correct, or would it make more sense to use

     token68        = 1*( ALPHA / DIGIT /
                          "-" / "." / "_" / "~" / "+" / "/" ) *"="

as a definition, and include those trailing padding characters?

I would treat the whole token (possibly including padding) as an opaque string, sha256-hash it, and do a db lookup based on the hash value.

tomhughes:
Is there any reason not to just take any string of non-whitespace characters? Why try and be clever about guessing what the token will look like? Unless the spec puts some limits on it, in which case follow those...

mmd-osm:
Right, the current implementation is too much following doorkeeper implementation details rather than the OAuth2 RFCs. https://datatracker.ietf.org/doc/html/rfc6750#section-2.1 includes "=* as additional trailing padding character(s), which I still need to add now.

mmd-osm:
Bearer token check now reflects the RFC I mentioned before. I will probably do another round of code review, as that frequently turns up interesting things.

By the way, https://oauth2.apis.dev.openstreetmap.org/ seems to be unavailable at the moment due to an Internal Server Error.

mmd-osm:
PPA publishing is still ongoing, and should be done real soon: https://launchpad.net/~mmd-osm/+archive/ubuntu/cgimap

mmd-osm:
It's a bit of a funny idea to roll out changes without doing any sort of database migration first. In this case it doesn't hurt because I moved all prepared statements to the place where they are actually used, rather than when the db connection is being set up, as it used to be the default approach with CGImap in the past.

tomhughes:
I don't believe I've said what my deployment plan is other than that I'd like to have all the pieces available so I can deploy them more or less simultaneously...

That said if it's safe to update cgimap first then I'll probably do that.

mmd-osm:
Well, the update already happened by some Chef magic behind the scenes. But there's no reason to worry. Like I said, OAuth2 is in some sleeping state right now, and all what happens right now is that you get a HTTP 503 when using "Bearer ..." authorization header.

tomhughes:
It must have been marked as a security update then - those deploy automatically overnight.

tomhughes:
Doesn't look like it was that but it's not clear what did trigger it - it has actually only happened on the three machines in Amsterdam and not it on the two in York, and at quite different times.

Needs more investigation, but it wasn't intended.

tomhughes:
Apparently chef was set to automatically upgrade but I don't think that's a good idea, especially as the timing depends on when each machine refreshes it's package metadata so they wind up spread out, so I have disabled that now.

mmd-osm:
I'm not exactly sure that the commit https://github.com/openstreetmap/openstreetmap-website/commit/e9010306c5b4970edf5cb8164feb43719d3bb63a did, but I believe it broke CGImap again... :(

tomhughes:
Ah yes sorry about that but I guess it will...

Existing tokens should still work but new tokens will now have the code stored as plain text instead of hashed with SHA256 so the token verification in cgimap will need to change :-(

I changed it because when I started trying to switch the web site to use OAuth 2 for it's notes API usage I realised that there was no way to reuse an existing token so every page load of the front page would have to generate a new one and we'd soon have millions of them. Likewise for iD once it switches over.

So I decided I needed to quickly change it before too many people got stuck in but I forgot it would affect cgimap as well.

In theory the rails code looks for a plain text token first, then a hashed one so that existing tokens continue to work.

PR #279, opened 2022-11-22
Replace deprecated std::iterator with direct member type declarations

Woazboat:
The [std::iterator](https://en.cppreference.com/w/cpp/iterator/iterator) helper class has been deprecated in C++ 17 in favor of direct declaration of the iterator member types. This causes the compiler to emit a ton of deprecation warnings, once for every source file that somehow includes the affected `infix_ostream_iterator.hpp` header.

* Fixed deprecation warnings by directly declaring the required member types and no longer deriving from std::iterator.
* `difference_type` changed to `std::ptrdiff_t` instead of `void` for [C++ 20 compatibility](https://en.cppreference.com/w/cpp/iterator/ostream_iterator).
* Add `iterator_concept` member type for [C++ 20 compatibility](https://en.cppreference.com/w/cpp/iterator/iterator_tags)

### DIFF
--- a/include/cgimap/infix_ostream_iterator.hpp
+++ b/include/cgimap/infix_ostream_iterator.hpp
@@ -4,18 +4,24 @@
 #include <ios>
 #include <ostream>
 #include <iterator>
+#include <cstddef>
 
 /**
  * output iterator which infixes its delimiter string between the output items.
  */
 template <typename _Tp, typename _CharT = char,
           typename _Traits = std::char_traits<_CharT> >
-class infix_ostream_iterator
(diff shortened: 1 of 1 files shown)


mmd-osm:
Thanks! I was already looking into replacing infix_ostream_iterator by fmt::join. As this has fairly low priority, I'll leave this PR open as a reminder, but will likely not merge it. Definitely not a topic for 2022 anymore.

mmd-osm:
I’ll leave the fmt::join topic for later, and merged this in the meantime.

Issue #281, opened 2022-11-24
Incorrect vector construction in pgsql selection of historic nodes/ways/relations


Woazboat:
When selecting historic nodes/ways/relations from the pgsql database, the id and version vectors are created via

```c++
  std::vector<osm_nwr_id_t> ids(eds.size());
  std::vector<osm_version_t> vers(eds.size());

  for (const auto &ed : eds) {
    ids.emplace_back(ed.first);
    vers.emplace_back(ed.second);
  }
```
https://github.com/zerebubuth/openstreetmap-cgimap/blob/master/src/backend/apidb/readonly_pgsql_selection.cpp#L577-L578

Presumably the vector is intended to be constructed with an appropriated pre-allocated/reserved size, but unfortunately the used `vector( size_type count );` constructor actually fills the vector with `count` default inserted elements. That means the `ids`/`vers` vectors are created with `eds.size()` elements with value `0` and then the real id/version elements are inserted in addition.
`reserve(eds.size())` should be used instead ([which is actually how it used to be previously](https://github.com/zerebubuth/openstreetmap-cgimap/commit/a12de48c8f010ae5e2e81f0a17daa56c4b50a93a#diff-58bbf3643eb7e8ee21dba4bd0b59e11210a37f3bcb28ad2f409c062b09ea305bL385))

https://en.cppreference.com/w/cpp/container/vector/vector (4)

mmd-osm:
Thanks, good catch, this should have been reserve instead, similar issues further down the in the same file. As we don’t have object / version 0 in the database, it won’t change the result.

PR #282, opened 2022-11-25
Fix vector reserve in pqsql select for historic n/w/r

Woazboat:
Fixes https://github.com/zerebubuth/openstreetmap-cgimap/issues/281


mmd-osm:
Thanks again for fixing the issue!

PR #284, opened 2022-12-25
Check user state, only allow db changes for confirmed and active users

### LINKED ISSUE #283: cgimap should check the user state
We should make cgimap check that the user state is `active` or `confirmed` and ignore any authentication for invalid users, either rejecting the request or processing it as if the user was unauthenticated for read requests.

mmd-osm:
Fixes #283

### DIFF
--- a/src/process_request.cpp
+++ b/src/process_request.cpp
@@ -34,10 +34,18 @@ void validate_user_db_update_permission (
   if (!user_id)
     throw http::unauthorized ("User is not authorized");
 
-  if (selection.supports_user_details ()
-      && selection.is_user_blocked (*user_id))
-    throw http::forbidden (
-	"Your access to the API has been blocked. Please log-in to the web interface to find out more.");
+  if (selection.supports_user_details ())
+  {
+    if (selection.is_user_blocked (*user_id))
+      throw http::forbidden (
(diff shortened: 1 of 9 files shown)


src/process_request.cpp
```diff
@@ -34,10 +34,18 @@ void validate_user_db_update_permission (
...
+  {
+    if (selection.is_user_blocked (*user_id))
+      throw http::forbidden (
+	  "Your access to the API has been blocked. Please log-in to the web interface to find out more.");
+
+    // User status has to be either active or confirmed, otherwise the request
+    // to change the database will be rejected
+    if (!selection.is_user_active(*user_id))
+      throw http::forbidden (
+          "You have not permitted the application access to this facility");
```
tomhughes, line 47: That's not really an accurate description is it? This would occur when there was an authorization but the user account had been closed or suspended.

mmd-osm: That’s exactly what Rails returned in that case, I’m only trying to replicate the same response here. I agree, the description isn’t all that helpful.

mmd-osm: api_controller.rb deny_access is where this error message originates from:

```ruby
  def deny_access(_exception)
    if doorkeeper_token || current_token
      set_locale
      report_error t("oauth.permissions.missing"), :forbidden
```

PR #289, opened 2023-09-12
Add ids to changeset comments output

AntonKhorev:
See https://github.com/openstreetmap/openstreetmap-website/pull/4245


mmd-osm:
Thanks! I'm merging this now so that other people can do a bit of additional testing on the dev instance.

By the way, did you plan any other changes in the near future? Since releasing new versions is always a bit of an effort, I'd like to keep the number of new releases as low as possible.

AntonKhorev:
I'm not going to change anything here probably at least for a month.

mmd-osm:
Hmm, good point. We already have some moderator role specific code in CGImap to handle redaction use cases (process_request.cpp, line 540), sot that's not a completely not concept. Maybe start the discussion again on Rails port. Once we have some more clarity, how exactly the behavior should look like for moderators (unconditionally show everything, or require some additional url parameter), we can revisit this issue here.

I started adding some cs comments on the dev instance: https://master.apis.dev.openstreetmap.org/changeset/273816
The new ideas also show up in the API response: https://master.apis.dev.openstreetmap.org/api/0.6/changeset/273816?include_discussion=true

AntonKhorev:
I tried adding a parameter for a different thing that is not in cgimap, got no response
https://github.com/openstreetmap/openstreetmap-website/pull/3934

mmd-osm:
I think getting no response is mostly a matter of having too many open issues and pull requests, and at the same time rather limited maintainer’s bandwidth. I would be ok to unconditionally show all comments for moderators. If they don’t like what they see, they can still fall back to a second non-moderator account.

AntonKhorev:
> I would be ok to unconditionally show all comments for moderators.
> If they don’t like what they see, they can still fall back to a second non-moderator account.

And maybe they won't like it, if it's hundreds of comments with swearing. But switching to another account may be inconvenient if an app that accesses the api is using some typical osm login library that supports only one active login. Omitting the auth token may also no work if gdpr api restrictions are implemented.

> One issue I see is that we don’t include any status information, I.e. it would be impossible for a moderator to figure out if a changeset comment is visible or hidden.

I'll add `visible=true/false` attribute like elements have. Although elements have separate deletion and redaction states and hiding comments is maybe more similar to redacting while their visible attribute corresponds to deleting.

Another question is whether it's worth showing this attribute to regular users and/or when hidden comments are not requested. All of the comments are going to be visible in this case. But maybe it's worth including some output for a hidden comment when requested. Of course for non-moderators it won't include the comment text and maybe won't include the author. But it's probably safe to show id. Hiding the date would be pointless because you can get a good estimate by looking at comments with id-1 and id+1. Regular users would see something like this if they request hidden comments:

```
<comment id="123" date="2023-09-19T19:33:11Z" visible="false" />
```

Then they'll know if there was a comment. Maybe there's a visible reply to that comment, now it wouldn't look odd.

AntonKhorev:
But the comments will be correctly sorted inside the `<discussion>`. Then with a hypothetical show-hidden-comments request you'll see something like:

```
<comment id="123" visible="true" ...><text>...</text></comment>
<comment id="122" visible="false"></comment>
<comment id="124" visible="true" ...><text>...</text></comment>
```

You'll know that the second comment is hidden even if ids are out of sequence, and that there was only one comment.

(reply to the original comment, you're still editing it?)

mmd-osm:
I was typing on my mobile, which is a bit inconvenient and I had to update my text a few times. What I had in mind was the current scenario, where we don't show any "visible=false" comments. Based on the following response, you wouldn't be able to figure out, if comments 124..139 were hidden, or if they belong to some other changeset.

```
<comment id="123" visible="true" ...><text>...</text></comment>
...
<comment id="140" visible="true" ...><text>...</text></comment>
```

I would expect changeset comment ids to be increasing. However, I don't recall how this is implemented on Rails, and if backward jumps from 123 to 122 are possible. Ideally, there would be some sort of lock on the changeset, then a new changeset comment id is generated, and finally a commit the new changeset comment.

AntonKhorev:
> Based on the following response, you wouldn't be able to figure out, if comments 124..139 were hidden, or if they belong to some other changeset.

Sure, even if I check all changesets, I might discover that those comments don't exist in any of them. They are hidden, but I don't know to which changeset they belong.

That's a reason to show them on request, omitting whatever is intended to be hidden. But it can't be done by default because no one yet expects comments with omitted attributes/text. If a parameter is added then it's going to be backwards compatible.

PR #294, opened 2023-10-28
Fix #290: Add moderators specific rate limiting settings

### LINKED ISSUE #290: Increase (or remove) the rate limit for administrators/moderators
Currently, [Thanos](https://github.com/Zaczero/Thanos) *(tool used by DWG)* encounters difficulties with mass-reverts due to rapidly reaching [the rate limit](https://github.com/openstreetmap/cgimap/blob/704c171bccf9f6a41811dc076cd9c405e45ab463/src/process_request.cpp#L511). This results in considerable delays for extensive revert tasks, sometimes requiring several hours of waiting. Ideally, the rate limit behavior should be adjusted for OSM moderation purposes.

DavidKarlas:
This change removes transfer size rate limiting for moderator role users aka. DWG members.

I tried to make change as small as possible and not refactor much code.

User roles were fetch before to check if user can see redactions or not...


tomhughes:
I envisaged an ability to set a separate, higher, limit for moderators not to remove it all together.

Also this will need changing on the rails side as well to be consistent.

DavidKarlas:
Ok, I will update PR, should we go with fixed value like 5x or 10x, or should I introduce `--moderatormultiplier`?
Looks like Ruby side doesn't have rate limiting at all, do you want me to add it?

tomhughes:
I would suggest just having a second set of limit and burst values for moderators.

DavidKarlas:
Pushed changes, added `--moderator-maxdebt` and `--moderator-ratelimit`.

DavidKarlas:
While working on this, I realized we are walking near limit which is 2GB(I set for moderators default to 1GB, which is 4x of previous default). How do we avoid problems? I see three options:
* Switch to `long`
* throw exception if option value is bigger than 1.5GB
* Change to `uint` and throw exception if option value is bigger than 3.5GB

src/rate_limiter.cpp
```diff
@@ -65,6 +77,7 @@ bool memcached_rate_limiter::check(const std::string &key) {
   memcached_return error;
 
   mc_key = "cgimap:" + key;
+  int bytes_per_sec = moderator ? moderator_bytes_per_sec : standard_bytes_per_sec;
```
mmd-osm, line 81: Note on refactoring: we should probably move parsing of parameters, and setting of default values to the central location for options handling in options.[ch]pp. 

Then use something like `global_settings::get_bytes_per_sec(moderator)` and `global_settings::get_max_bytes(moderator)`  to fetch the relevant limits for the current user type.

mmd-osm:
> Switch to long(~doubles size of memcached memory)

I see another issue here. Since we're storing values in binary format in memcached, any change to the length would either require to purge the cache, or we need code to deal with the previous 32 bit length, and a new 64 bit length.
Also, I'm not exactly clear what Rails is doing with rate limiting.

Then the rate limiting code is a bit dated, and test coverage isn't exactly perfect. Maybe we should think about how we could add some test for the new code.

tomhughes:
I think I was wrong about rails actually... I don't think we actually implement the rate limiting there.

src/process_request.cpp
```diff
@@ -516,7 +516,7 @@ void process_request(request &req, rate_limiter &limiter,
     }
 
     // check whether the client is being rate limited
-    if (!limiter.check(client_key)) {
+    if (!limiter.check(client_key, user_roles.count(osm_user_role_t::moderator) > 0)) {
```
mmd-osm, line 519: Another hint on refactoring: we have the same code in line 583. Maybe introduce a `const bool is_moderator = ... `and use that in both places.

DavidKarlas: Fixed.

src/rate_limiter.cpp
DavidKarlas: Done.

src/options.cpp
```diff
@@ -138,6 +142,46 @@ void global_settings_via_options::set_oauth_10_support(const po::variables_map &
   }
 }
 
+void global_settings_via_options::set_bytes_per_sec(const po::variables_map &options) {
```
mmd-osm, line 145: Seems like I commented on this in your fork, which is why it doesn't show up in your pull request.

You need to call both new setter methods from global_settings_via_options::set_new_options, line 43 in the same file, otherwise we're not updating our option values with the externally provided values.

DavidKarlas: Didn't have time to test changes, will fix & test it tonight

DavidKarlas:
I will attempt to mock now() to unit test rate limitting.

DavidKarlas:
I wonder if I should rename:

`global_settings::get_bytes_per_sec(moderator);`
into
`global_settings::get_ratelimiter_ratelimit(moderator);`

and

`global_settings::get_max_bytes(moderator);`
into
`global_settings::get_ratelimiter_maxdebt(moderator);`
?

mmd-osm:
Yes, I think that's a good idea. I was also a bit unhappy with the naming, because it's difficult to see that both a related to rate limiting.

Zaczero:
BTW, do you accept donations, @DavidKarlas? I really appreciate it when someone goes out of their way to help others.

DavidKarlas:
> BTW, do you accept donations, @DavidKarlas? I really appreciate it when someone goes out of their way to help others.

[Go here](https://supporting.openstreetmap.org/donate/) 😉
 
I pushed now renamed global settings and actually do parsing of options...
I tested this, checked via debugger that options are correctly applied, observed via JOSM that limits are never reached for moderator user... I'm pretty confident this change is good...

Now question... If we want to add test for this, I need guidance on how to mock `time(NULL)`
As far as I can see there are 2 ways, would like to hear more...
1) I introduce new class for getting time that similar to the way null_rate_limiter vs. memcached_rate_limiter, this means that in `main.cpp` this would be created and passed to `process_request(req, limiter, generator, route, *factory, update_factory.get(), oauth_store.get())`
2) Add time parameter to `rate_limiter.check` and `rate_limiter.update` that is `epoch` seconds and only unit tests `rate_limiter` and not rest of `process_request` logic like passing in correct `bytes_written` value...
3) Third option?

If you want test let me know how to mock, or if you think this can be merged without test please review.
Thank you!

Zaczero:
> [Go here](https://supporting.openstreetmap.org/donate/) 😉

2nd time in a week? Give me a break!

DavidKarlas:
I wonder what would be implications of using `req.get_current_time()` which is set here: https://github.com/zerebubuth/openstreetmap-cgimap/blob/5b5f18b6a6c76d198a59db1ddbfc8889fa14ba47/src/main.cpp#L206-L207
This would allow for easy mocking/calling of `process_request` similar to https://github.com/zerebubuth/openstreetmap-cgimap/blob/5b5f18b6a6c76d198a59db1ddbfc8889fa14ba47/test/test_core.cpp#L484

mmd-osm:
> If you want test let me know how to mock, or if you think this can be merged without test please review.

I think I would need some time for a good idea how to test this. As mentioned earlier, I would be ok to do this later as a separate PR in this case.

src/options.cpp
```diff
@@ -138,6 +144,46 @@ void global_settings_via_options::set_oauth_10_support(const po::variables_map &
...
+  }
+}
+
+void global_settings_via_options::set_ratelimiter_maxdebt(const po::variables_map &options) {
+ if (options.count("maxdebt")) {
+    auto parsed_max_bytes = options["maxdebt"].as<long>();
+    if (parsed_max_bytes < 0)
+      throw std::invalid_argument("maxdebt must be a positive number");
+    if (parsed_max_bytes > 3500)
+      throw std::invalid_argument("maxdebt must be 3500 MB or less.");
```
mmd-osm, line 173: This could confuse some folks, since it looks similar to the error message for ratelimit. However, we're actually looking for a value <= 3500.

Maybe: `maxdebt must be 3500 or less (unit: MB)`.

mmd-osm: That's pretty much the only thing I still have on my list. I would continue with merging afterwards. Hopefully we could give it a try on the dev instance.

DavidKarlas: Updated

mmd-osm: Great, thanks a lot! I will merge right after the two pipeline checks are done.

PR #295, opened 2023-10-29
Fix #291: Add Retry-After http header in 509 response

### LINKED ISSUE #291: Add Retry-After http header in 509 response
Error response could include some details about the expected wait time in seconds.
https://www.rfc-editor.org/rfc/rfc9110.html#name-retry-after

DavidKarlas:
Let me know how ugly solution it is from 0 to 10 to do `if (e.code() == 509)` and then `dynamic_cast`. It shows that my language is C# :P

### DIFF
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -147,8 +147,8 @@ void global_settings_via_options::set_oauth_10_support(const po::variables_map &
 void global_settings_via_options::set_ratelimiter_ratelimit(const po::variables_map &options) {
   if (options.count("ratelimit")) {
     auto parsed_bytes_per_sec = options["ratelimit"].as<long>();
-    if (parsed_bytes_per_sec < 0)
-      throw std::invalid_argument("ratelimit must be a positive number");
+    if (parsed_bytes_per_sec <= 0)
+      throw std::invalid_argument("ratelimit must be greater than zero");
     if (parsed_bytes_per_sec > 1024 * 1024 * 1024)
       throw std::invalid_argument("ratelimit must be 1GB or less");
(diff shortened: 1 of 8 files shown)


tomhughes:
It theory it's safe to use `static_cast` there because we know from the code check that the object should be of that type.

Now you can argue that `dynamic_cast` verifies that but as this is a reference it does so by throwing an exception though I guess that just means this request will abort - the alternative would be to do a pointer case instead and then just not add the header if you get null.

DavidKarlas:
Did dynamic_cast of pointer, and added if, to not crash, thank you!
Fixed unit tests, and reverted change of 404 return type from text/plain to text/html because tests started failing... Hence I think it should be separate PR...

src/options.cpp
```diff
@@ -147,17 +147,17 @@ void global_settings_via_options::set_oauth_10_support(const po::variables_map &
 void global_settings_via_options::set_ratelimiter_ratelimit(const po::variables_map &options) {
   if (options.count("ratelimit")) {
     auto parsed_bytes_per_sec = options["ratelimit"].as<long>();
-    if (parsed_bytes_per_sec < 0)
-      throw std::invalid_argument("ratelimit must be a positive number");
+    if (parsed_bytes_per_sec <= 0)
+      throw std::invalid_argument("ratelimit must be greater than zero");
```
mmd-osm, line 151: Good catch. I thought about some division by zero issue in rate_limiter.cpp, but later forgot about it again.

mmd-osm:
Thanks! Looks good to me, let's give it a try and merge!

PR #300, opened 2023-11-07
Add create single node endpoint

AntonKhorev:
Currently cgimap is missing single-element write api calls. If all write calls were in one place, maybe it would be easier to implement rate limits and other checks for uploaded data?

Here's an attempt to add `PUT node/create`. I started doing it before the new user rate limiter code was added. Not sure how compatible it is with the latest changes. Is it a good idea at all?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@ test/.deps/
 test/.libs/
 test/test_core
 test/test_apidb_backend_nodes
+test/test_apidb_backend_node_changes
 test/test_apidb_backend_oauth
 test/test_apidb_backend_oauth2
 test/test_apidb_backend_historic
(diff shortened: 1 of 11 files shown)


mmd-osm:
Thanks a lot for looking into this. I also thought about adding this a few years back, but in the end decided not to implement it. I'd rather suggest to deprecate these single object endpoints on the Rails port, and in the end remove them altogether. Of course, that's a discussion we need to start on the Rails repo first.

So regarding the rate limiting, we have a common implementation in place both for Rails and CGImap, so that's nothing to worry about. We have already fully addressed this issue. No matter if you're uploading single objects through Rails or uploading large changes through CGImap, the end result will be exactly the same.

pnorman:
Note that it's probably going to be a lot harder to hit rate limits through single object uploads because their design is inherently slower.

mmd-osm:
Right, single object upload is significantly slower,  a 5000x slowdown isn't totally unexpected. Performance is obviously another topic why I don't want to spend much time on the single upload, besides that you can't upload multiple changes in a single db transaction. It's a left over from pre API 0.6 times, and I believe it's largely obsolete these days.

AntonKhorev:
> I'd rather suggest to deprecate these single object endpoints on the Rails port, and at some point remove them altogether.

There are users who wouldn't care if these points are deprecated as long as they are usable.

> Regarding the rate limiting, we have a common implementation in place both for Rails and CGImap, so that's nothing to worry about. We have already fully addressed this issue.

The wider issue is people willing to damage the data. It's impossible to fully address.

PR #302, opened 2023-11-11
Adding right text for HTTP "Status" field 

mmd-osm:
Status is a non-standard http header field, but we should show the right text for http 429.

### DIFF
--- a/src/request_helpers.cpp
+++ b/src/request_helpers.cpp
@@ -103,6 +103,7 @@ const char *http_message_status_410 = "Gone";
 const char *http_message_status_412 = "Precondition Failed";
 const char *http_message_status_413 = "Payload Too Large";
 const char *http_message_status_415 = "Unsupported Media Type";
+const char *http_message_status_429 = "Too Many Requests";
 const char *http_message_status_509 = "Bandwidth Limit Exceeded";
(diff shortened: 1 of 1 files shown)


tomhughes:
Actually `Status` is standard in HTTP/2 I believe, as a replacement for the message text part of the initial response line which no longer exists.

mmd-osm:
zerebubuth added this Status header back in Feb 2015 in https://github.com/zerebubuth/openstreetmap-cgimap/commit/e90291bd1fd3f3f0ea3389d7b04d8ce78afc8e63 

HTTP/2 was introduced a few months later in May 2015, and  I'm not entirely clear if the Status header has been introduced as to anticipate any future HTTP/2 standard. It's not completely impossible, though the commit message doesn't mention HTTP/2 explicitly. Then we also have FastCGI in the picture, which might do some funny stuff on its own.

Am I right in saying that the status code would be the only thing that is relevant for HTTP/2 here? I haven't really seen any issue before, and only did this pull request for consistency purposes.

tomhughes:
I think I was misremembering... It doesn't have a response line at all but does have a `:status` pseudo-header that carries the numeric code only but not the message.

Something (possibly rails) adds a `Status` header to responses from the main site that echos the normal response line.

PR #304, opened 2023-11-13
Cleanup & code de-duplication

Woazboat:
Some cleanup, code de-duplication and refactoring. See commits for details.

### DIFF
--- a/include/cgimap/api06/handler_utils.hpp
+++ b/include/cgimap/api06/handler_utils.hpp
@@ -5,12 +5,12 @@
 #include "cgimap/request.hpp"
 #include "cgimap/api06/id_version.hpp"
 #include <vector>
-#include <string>
+#include <string_view>
 
 namespace api06 {
 
 std::vector<id_version> parse_id_list_params(request &req,
-                                             const std::string &param_name);
+                                             std::string_view param_name);
 }
 
 #endif /* API06_HANDLER_UTILS_HPP */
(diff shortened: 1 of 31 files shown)


mmd-osm:
Review status:

* [ ] 134aa31702df6e248928fd18fb31bd1189e67420 Use string_view in parse_id_list_params()
* [ ] dfff59ba9a3e3e982311f6c86d9eba6c43597a8e Deduplicate http header formatting and move to http file
* [ ] ec13ff2cf8d1b0e00c1a3b9ec7ed1eff66bb6e94 Move headers_t definition to http header file
* [ ] 16c7f42063425de9850cba866a76a9cf9d331a32 Check for empty callback function
* [ ] 20187e65911afbc9a9081a90458852ac9cc27b46 Use structured binding for http headers
* [ ] b225e015ac66390ee8ce802711fda2d7b57a4c9a Minor cleanup improvements and formatting
* [ ] f663bb2725994715bdbf96335b35cc297de2c876 Also reset allowed methods when resetting request
* [ ] a3dc7061686bf93b51840911bb993a58266dc5e2 Deduce underlying type for method enum
* [ ] ab6138f78233df0b5c2e2bfb991331d10697574f Use structured binding for historic object id/version
* [ ] d1654ee0fcf3ddc2ddf88afcb1c2f33b5d055235 Simplify constructors, use in-class initializers
* [ ] cb5c034e17ad21a8338648161b0b7abc67b56197 Add override/const
* [ ] 65c6612ea6f4cc40250b21365fb628d94ac865f7 Deduplicate http status message strings

mmd-osm:
Thanks a lot for this PR. I have reviewed all commits as per the above list, and this looks fine to me.

One thing I've noticed is that in some cases you have moved a bit of code from fcgi_request.cpp to http.cpp, which means that it would end up in a different shared library. I don't see an immediate issue with that, since the cgimap shared library concept isn't really used productively.

Woazboat:
I think it was from request_helpers.cpp to http.cpp which are both in libcgimap_core

PR #305, opened 2023-11-16
Use catch2 unit test framework

Woazboat:
Replace ad-hoc tests with catch2 unit test framework. (Only `test_parse_osmchange_input` for now.)
- Tests are much easier to write/maintain
- More information about test failures
- Tests can be selectively run, randomised, ...
- ...

The CI using the ubuntu container currently fails because it uses an old version 20.04 where catch2 is not available in the package repository (it's incorrectly labeled as ubuntu 22.04, where catch2 _would_ be available).


mmd-osm:
I think it makes a lot of sense to get rid of the home grown ad hoc test approach, and use a proper test framework for unit tests . Back in the days, I used the existing way of writing tests as to avoid introducing too many changes at once.

I've fixed the inconsistent naming for the Ubuntu 20.04 CI run, and also added a dedicated CI run for Ubuntu 22.04. Maybe you can try to rebase this pull request to master to see if this works better now.

I'm a bit undecided if we should keep 20.04 support around, or add catch2 as a contrib library instead (like here: https://github.com/osm2pgsql-dev/osm2pgsql/tree/master/contrib/catch2). For osm.org, we only need 22.04, but other folks might run some older Ubuntu versions still.

Woazboat:
Yeah, it was a bit of a hassle to try and find whatever the next test number should be when adding a new test. 

I've rebased the PR, but apparently debian testing decided to switch to catch2 3.0 at this very moment...
I'll need to check if changes are needed since it's a bit different

Woazboat:
Okay, so I think adding catch2 2.13.7 as a contrib library is definitely the easiest way forward. catch2 2.x and 3.x are not really compatible and require some bigger changes to the build system since it's no longer a header only library.
Making it compatible for both 2.x and 3.x isn't really possible and distributions are going to be a mix of both for a while.

mmd-osm:
I agree, with all the complications around different catch2 versions across distributions and the required build system changes for 3.x, I'd say, let's go for the 2.13.7 contrib option.

mmd-osm:
By the way, I also added a CI run for Debian Bookworm in the meantime, since OSM operations plans to switch to Debian  next year (https://en.osm.town/@Firefishy/111428439382240705).

Woazboat:
Did some git juggling to add the relevant files of catch2 2.13.10 as a partial git subtree and to clean up the commit history for the PR. 

Github doesn't display it properly, but this is the commit log graph for this PR:
```
1c558679 * Convert parse_osmchange_input test to catch2
3a4b8495 * Add README.contrib for catch2
2c91156b * Add catch2 contrib lib as squashed + pruned git subtree
         |\  
d165e424 | * Squashed 'contrib/catch2/' content from commit 182c910b
b0d37f32 * CI: Added Debian Bookworm
```

test/test_parse_osmchange_input.cpp
```diff
@@ -481,401 +255,221 @@ void test_node() {
...
+TEST_CASE("Create relation, member ref too large", "[osmchange][relation][xml]") {
+  REQUIRE_THROWS_AS(process_testmsg(
+    R"(<osmChange><create><relation changeset="972" id="-1">
+           <member type="relation" ref="99999999999999999999999999999999" role="stop"/>
+           </relation></create></osmChange>)"), http::bad_request);
+}
 
-  // Member: role may have up to 255 unicode characters
-  for (int i = 1; i <= 256; i++) {
+TEST_CASE("Create relation, role with <= 255 unicode characters", "[osmchange][node][xml]") {
```
mmd-osm, line 440: There might be a typo/copy&paste issue here. I would expect `[osmchange][relation][xml]`

test/test_parse_osmchange_input.cpp
```diff
@@ -481,401 +255,221 @@ void test_node() {
...
-      if (i <= 255)
-        throw std::runtime_error("test_relation::010: Unexpected exception for "
-                                 "string length <= 255 characters");
-    }
+           v)));
   }
 }
 
-void test_invalid_data() {
+TEST_CASE("Create relation, role with > 255 unicode characters", "[osmchange][node][xml]") {
```
mmd-osm, line 452: Also here:  `[osmchange][relation][xml]`

test/test_parse_osmchange_input.cpp
```diff
@@ -481,401 +255,221 @@ void test_node() {
...
+
+TEST_CASE("Invalid data", "[osmchange][relation][xml]") {
+  REQUIRE_THROWS_AS(process_testmsg("\x3C\x00\x00\x00\x00\x0A\x01\x00"), http::bad_request);
 }
 
-void test_large_message() {
+
+// LARGE MESSAGE TESTS
+
+TEST_CASE("Very large XML message", "[osmchange][relation][xml]") {
```
mmd-osm, line 472: Very large XML message is testing nodes only, so maybe the tag should be `[osmchange][node][xml]`?

test/test_parse_osmchange_input.cpp
```diff
@@ -906,99 +500,69 @@ void test_large_message() {
...
+  REQUIRE_NOTHROW(process_testmsg(s.str()));
 }
 
-void test_object_limits() {
 
+// OBJECT LIMIT TESTS
+
+TEST_CASE("Create node, tags < max tags", "[osmchange][node][xml]") {
   auto test_settings = std::unique_ptr<global_settings_test_class>(new global_settings_test_class());
   test_settings->m_element_max_tags = 5000;
```
mmd-osm, line 511: By the way, thanks to catch2 it's fairly obvious now that this test case is by far the slowest one in this file:

```
test/test_parse_osmchange_input -D 1
7.754 s: Create node, tags < max tags
===============================================================================
All tests passed (7859 assertions in 73 test cases)

```

mmd-osm:
Overall this looks really good, also the git juggling is much appreciated, much better than introducing submodules.
I only added a few remarks regarding tagging of test cases. All of them are fairly minor.

Woazboat:
Yeah, those were copy&paste errors. I fixed them now

PR #316, opened 2023-11-27
Build system/include cleanup + some more tests

Woazboat:
- Fix includes 
- Modify preprocessor conditions for HAVE_YAJL to also respect HAVE_YAJL=0
- Move psql code to apidb lib
- ~~Allow path to structure.sql database schema for tests to be configured via command line arguments~~
- Some more tests


src/data_selection.cpp
```diff
@@ -1,55 +1,4 @@
 #include "cgimap/data_selection.hpp"
 
 
-std::vector<std::string> psql_array_to_vector(std::string str) {
```
mmd-osm, line 4: data_selection.cpp seems to be rather empty now. Do we still need it?

test/test_apidb_backend_changesets.cpp
```diff
@@ -857,10 +859,26 @@ void test_changeset_close(test_database &tdb) {
 
 } // anonymous namespace
 
-int main(int, char **) {
+int main(int argc, char** argv) {
```
mmd-osm, line 862: Somehow this introduces a bit of code duplication across test cases. I do realize that this would be useful for cmake. Since it's not needed right now for the current autotools based build, I'd like to keep this commit separate from the others. Either I could cherry-pick all other commits and omit this one, or you could split this up on your end. Whatever works best for you.

Woazboat: Yeah, it could probably be refactored. I removed it from this PR for now.

test/test_parse_osmchange_input.cpp
```diff
@@ -412,6 +424,30 @@ TEST_CASE("Create way, node ref missing", "[osmchange][way][xml]") {
...
+
+TEST_CASE("Delete way, no id", "[osmchange][way][xml]") {
+  REQUIRE_THROWS_AS(process_testmsg(
+    R"(<osmChange><delete><way changeset="858" version="1"/></delete></osmChange>)"),
+    http::bad_request);
+}
+
+TEST_CASE("Delete way, no changeset", "[osmchange][way][xml]") {
+  REQUIRE_THROWS_AS(process_testmsg(
+    R"(<osmChange><delete><way id="-1 version="1"/></delete></osmChange>)"),
```
mmd-osm, line 441: I think there's a quotation mark missing right after `<way id="-1`  - similar issue in line 447 below

PR #318, opened 2023-11-28
Add more override, const, default

Woazboat:


include/cgimap/api06/changeset_upload/osmobject.hpp
```diff
@@ -97,9 +97,9 @@ namespace api06 {
 
     osm_nwr_signed_id_t id() const { return *m_id; }
 
-    bool has_changeset() const {  return ((m_changeset) ? true : false); }
-    bool has_id() const { return ((m_id) ? true : false); };
-    bool has_version() const { return ((m_version) ? true : false); }
+    bool has_changeset() const {  return (m_changeset ? true : false); }
```
mmd-osm, line 100: maybe we could use `m_changeset.has_value()`  and get rid of `? true : false`. Same for the next two lines.

include/cgimap/api06/changeset_upload/osmobject.hpp
```diff
@@ -147,7 +147,7 @@ namespace api06 {
 
     virtual std::string to_string() {
 
-      return fmt::format("{} {:d}", get_type_name(), ((m_id) ? *m_id : 0));
+      return fmt::format("{} {:d}", get_type_name(), (m_id ? *m_id : 0));
```
mmd-osm, line 150: How about `m_id.value_or(0)` ?

include/cgimap/infix_ostream_iterator.hpp
```diff
@@ -26,13 +26,13 @@ class infix_ostream_iterator {
 private:
   ostream_type *_M_stream;
   const _CharT *_M_string;
-  bool _M_first;
+  bool _M_first{true};
 
 public:
   infix_ostream_iterator(ostream_type &s)
-      : _M_stream(&s), _M_string(0), _M_first(true) {}
+      : _M_stream(&s), _M_string(0){}
```
mmd-osm, line 33: Add one space before {} to be in line with line 35 below?

mmd-osm:
Thanks! Just adding a few comments, otherwise LGTM.

include/cgimap/api06/changeset_upload/osmobject.hpp
Woazboat: yup, changed to `has_value()`

include/cgimap/infix_ostream_iterator.hpp
Woazboat: fixed

PR #332, opened 2023-12-05
Automatically create github release for tagged versions

Woazboat:
Use github actions to automatically create a release for versions tagged with `v*.*.*`

- Build a docker image and upload to github container repository.
- Upload pre-built binary as release artifact.

Currently based on Ubuntu 22.04 docker image

### DIFF
--- a/.github/workflows/tag_release.yml
+++ b/.github/workflows/tag_release.yml
@@ -1,0 +1,70 @@
+name: Create release for tagged versions
+
+on:
+  push:
+    tags:
+      - 'v*.*.*'
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+  BUILD_TYPE: Release
+
+jobs:
+  build:
+    name: Create release
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      packages: write
+
+    steps:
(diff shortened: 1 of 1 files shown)


mmd-osm:
As a heads up: This may probably not work since I’m only collaborator in this repo and cannot create personal access tokens or configure a GITHUB_TOKEN.

mmd-osm:
I think we need to double check this with zerebubuth, who owns the repo. While we might be able to publish packages, I have no way to control or manage them afterwards. 

Also, we should check if publishing packages is really worth the extra effort and does not increase the overall maintenance effort. osm.org uses Ubuntu PPAs at the moment, OpenHistoricalMap is building their own Docker images, so both of our main "customers" would not see an immediate benefit.

I introduced Docker images mainly to have a bit more control on package dependencies when running unit tests. They could be also useful for some local testing. However, in general are not meant as is for production usage. Once we start publishing packages, people might have a different expectation, i.e. would expect a more "production like" image.

mmd-osm:
Now with the discussion in #381 we could attach the build artifact to the new release, which would be a single .deb file in this case. This could avoid some of issues with a package registries for now.

mmd-osm:
I've included at least the softprops/action-gh-release based release process in #449 now. Thanks a lot for the PR!

PR #335, opened 2023-12-08
Update make clang-format target

Woazboat:
This at least allows it to run again. The `.clang-format` style file produces some suboptimal results and should be updated.


mmd-osm:
Thanks!

PR #336, opened 2023-12-08
Daemon mode

### LINKED ISSUE #330: Follow up on OpenHistoricalMap issue
Context: https://github.com/OpenHistoricalMap/issues/issues/642

Not running in daemon mode:
* cgimap binary starting up as a single process, handling one user at a time only.
* Not meant to be run in standalone mode: needs to be managed by supervisord or lighttpd, to ensure proper restarts in case of failure, and more importantly, start up multiple processes.
* CGImap  exits with non-zero return code in case of error.
* Remove parameter "--instances=30" in Dockerfile, since it's only relevant in daemon mode.
* Maybe add warning message during startup in this case.

Running in daemon mode:
* Automated restarting of forked processes works ok in case of database error.

Overall:
* Improve documentation, clarify how daemon / non-daemon modes behave.

mmd-osm:
List of changes:

* Show warning when ` --instances` command line parameter is set when in non-daemon mode

* Dockerfiles: To facilitate monitoring, logfiles are now sent to the Docker daemon's stdout, and are accessible using `docker logs -f ...` . Also, Dockerfiles are starting CGImap in daemon mode, to automatically restart forked CGImap processes. The container stops, if the main CGImap process terminates, or docker stop is invoked on the container.

* Dockerfile: Default number of instances in daemon mode is now 10 instead of 30.

* Some refactoring into non-daemon/daemon mode

Fixes #330

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,7 @@ COPY . ./
 # Compile, install and remove source
 RUN ./autogen.sh && \
     ./configure --enable-static --disable-shared --enable-yajl && \
-    make -j3 && \
+    make -j${nproc} check TESTS= && \
     make check || ( cat ./test-suite.log ; exit 1 ) && \
     strip openstreetmap-cgimap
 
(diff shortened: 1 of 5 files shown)


Woazboat:
> * Show warning when ` --instances` command line parameter is set when in non-daemon mode

This warning is now always shown in non-daemon mode, even when the `--instances` commandline option is not specified since it has a default value of 5.

mmd-osm:
Right, I will fix that in a minute, PR is on the way...

PR #338, opened 2023-12-10
Make signal flags atomic for explicit signal safety

Woazboat:
https://eel.is/c++draft/support.signal

A simple boolean will almost always work in practice, but technically it's not thread/signal safe and behavior depends on the compiler and architecture.


mmd-osm:
Good catch. It's a bit surprising that the code didn't even use the pre C++11 `sig_atomic_t volatile` in that case.

PR #340, opened 2023-12-10
Install signal handlers for daemon and non-daemon mode

Woazboat:
Install signal handlers for daemon and non-daemon mode + split into separate function

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -246,6 +246,26 @@ void process_requests(int socket, const po::variables_map &options) {
   req.dispose();
 }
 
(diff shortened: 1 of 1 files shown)


mmd-osm:
Good idea to make this a bit more consistent.

mmd-osm:
We’re getting some compiler warnings now, sa as a variable is no longer needed after refactoring…

#9 62.79 main.cpp: In function 'void {anonymous}::daemonise()':
#9 62.79 main.cpp:274:20: warning: unused variable 'sa' [-Wunused-variable]

PR #342, opened 2023-12-11
Reduce temporary allocations for xml serialization

Woazboat:
- Use std::to_chars to convert integers to strings for xml serialization instead of xml attribute string formatter 
  - No temporary heap allocations
    - At _least_ 4 temporary heap allocations per element eliminated (element id, element version, changeset number, last user id)
  - std::to_chars is generally the fastest string conversion function available
- Add char* overloads for xml_writer functions to avoid temporary std::string allocations
  - Passing string literals to a const std::string& parameter requires a temporary std::string object. (libxml2 only uses char* anyway.)
- Deduplicate and move common string constants to base formatter class

### DIFF
--- a/include/cgimap/xml_writer.hpp
+++ b/include/cgimap/xml_writer.hpp
@@ -10,12 +10,16 @@
 #ifndef WRITER_HPP
 #define WRITER_HPP
 
+#include "cgimap/output_buffer.hpp"
+#include "cgimap/output_writer.hpp"
+#include "cgimap/util.hpp"
+
 #include <memory>
 #include <string>
 #include <stdexcept>
-#include "cgimap/output_buffer.hpp"
-#include "cgimap/output_writer.hpp"
 #include <cinttypes>
+#include <charconv>
+#include <array>
 
 /**
  * Writes UTF-8 output to a file or stdout.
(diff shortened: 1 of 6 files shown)


Woazboat:
Before
Screenshot_20231211_133255
Screenshot_20231211_133729


After
Screenshot_20231211_133340
Screenshot_20231211_133940

include/cgimap/xml_writer.hpp
```diff
@@ -38,24 +42,40 @@ class xml_writer : public output_writer {
...
-  void attribute(const std::string &name, bool value);
+  void attribute(const char *name, double value);
+  void attribute(const char *name, bool value);
+
+  template<typename T>
+  inline void attribute(const std::string &name, T value) { attribute(name.c_str(), value); }
+
+  template<typename TInteger, std::enable_if_t<std::is_integral_v<TInteger>, bool> = true>
+  void attribute(const char *name, TInteger value) {
+    std::array<char, string_buf_size_v<TInteger>> buf;
```
mmd-osm, line 63: Could we maybe use some hardcoded value here instead, like 32? The array is residing on the stack only, and the memory isn't being initialized, so there shouldn't be much of an extra cost to have a larger buffer size. This way, we wouldn't need the additional logic in util.hpp. I believe this could be good enough for our purposes.

Woazboat: That works until someone passes in a `__int128`. (It's probably not very likely that will ever happen, but who knows.)

The `static_assert`s here https://github.com/zerebubuth/openstreetmap-cgimap/pull/342/files#diff-64a14965469222c18245822fae47ea2eff593b71d3a397d19e773bff633fae62R154-R159 provide some sanity checks that the size logic works as intended.

(`static_assert(string_buf_size_v<__int128> == 45);` for `__int128`)

mmd-osm:
I also checked what the impact of these changes would be on the runtime. For this purpose, I tried a /map call with and without your changes in place, both times without compression. I could see some 10-15% improvement on my (not super representative) test data set.

One noticeable point is the conversion of lat/lon double values to a fixed 7 digit representation, like for [null island](https://www.openstreetmap.org/api/0.6/node/3815077900). I believe this might be a bit faster when using fmt to generate the string.

image

Woazboat:
I would have liked to do some benchmarking as well but don't have a good setup for that yet. 

The attribute(..., double) conversion could definitely be sped up as well. Using fmt would very likely be faster than the current implementation. We could also use `std::to_chars` like for the integer attributes, which should be even faster. I didn't include that here because determining the required buffer size for floating point values isn't that straightforward, but we could also just use some large buffer size https://stackoverflow.com/questions/1701055/what-is-the-maximum-length-in-chars-needed-to-represent-any-double-value

I did a short quickbench comparison between `snprintf` (which is likely what libxml2 uses internally), `std::format` (because fmtlib isn't available in quickbench) and `std::to_chars`:
https://quick-bench.com/q/z8ozHON5FOmMyW6N00WvpmicaxA

`std::to_chars` is 40 times faster than `snprintf` and 3 times faster than `std::format`.

Woazboat:
Ugh, the old gcc in Ubuntu 20.04 doesn't have support for floating point `std::to_chars` 
https://en.cppreference.com/w/cpp/compiler_support/17 ([P0067R5](https://wg21.link/P0067R5))

Woazboat:
And also doesn't support precompiled fmt strings...

Woazboat:
https://www.zverovich.net/2020/06/13/fast-int-to-string-revisited.html

Woazboat:
> Then extract_tags / extract_elements might be slowed down a bit by psql_array_to_vector: the usage of streams may not be ideal here. --> https://cpp.godbolt.org/z/MsnGE9Ehv


Lets move that discussion over here: https://github.com/zerebubuth/openstreetmap-cgimap/pull/344

include/cgimap/xml_writer.hpp
```diff
@@ -38,24 +42,41 @@ class xml_writer : public output_writer {
...
 
   // overloaded versions of writeAttribute for convenience
-  void attribute(const std::string &name, double value);
-  void attribute(const std::string &name, int32_t value);
-  void attribute(const std::string &name, int64_t value);
-  void attribute(const std::string &name, uint32_t value);
-  void attribute(const std::string &name, uint64_t value);
-  void attribute(const std::string &name, bool value);
+  void attribute(const char *name, double value);
+  void attribute(const char *name, bool value);
```
mmd-osm, line 56: Since `std::is_integral_v<bool> == true`, I would somehow expect full template specialization for "bool" instead of an overloaded version: https://cpp.godbolt.org/z/M1GTev3jv

mmd-osm:
Thanks, also merging this now. I removed my previous comment on template specialization for attribute(name, bool), since overload resolution happens first.

PR #344, opened 2023-12-12
Use std::string / std::string_view for psql_array_to_vector

Woazboat:
std::string append is ~2-3 times faster than string streams

https://quick-bench.com/q/JLjFJX98msn1zUpsn0I6a0voUqc

### DIFF
--- a/include/cgimap/backend/apidb/common_pgsql_selection.hpp
+++ b/include/cgimap/backend/apidb/common_pgsql_selection.hpp
@@ -16,6 +16,9 @@
 #include <chrono>
 #include <functional>
 #include <pqxx/pqxx>
+#include <string_view>
+#include <string>
+#include <vector>
 
 /* these functions take the results of "rolled-up" queries where tags, way
  * nodes and relation members are aggregated per-row.
@@ -50,6 +53,6 @@ void extract_changesets(
 
 // parses psql array based on specs given
 // https://www.postgresql.org/docs/current/static/arrays.html#ARRAYS-IO
-std::vector<std::string> psql_array_to_vector(std::string str);
+std::vector<std::string> psql_array_to_vector(std::string_view str);
 
 #endif /* CGIMAP_BACKEND_APIDB_COMMON_PGSQL_SELECTION_HPP */
--- a/src/backend/apidb/common_pgsql_selection.cpp
+++ b/src/backend/apidb/common_pgsql_selection.cpp
@@ -283,54 +283,54 @@ void extract_changesets(
   }
 }
 
-std::vector<std::string> psql_array_to_vector(std::string str) {
+std::vector<std::string> psql_array_to_vector(std::string_view str) {
   std::vector<std::string> strs;
-  std::ostringstream value;
+  std::string value;
   bool quotedValue = false;
   bool escaped = false;
   bool write = false;
 
-  if (str == "{NULL}" || str == "")
+  if (str == "{NULL}" || str.empty())
     return strs;
 
-  for (unsigned int i=1; i<str.size(); i++) {
-    if (str[i]==',') {
+  const auto str_size = str.size();
+  for (unsigned int i = 1; i < str_size; i++) {
+    if (str[i] == ',') {
       if (quotedValue) {
-        value<<",";
+        value += ",";
       } else {
         write = true;
       }
-    } else if (str[i]=='\"') {
+    } else if (str[i] == '\"') {
       if (escaped) {
-        value << "\"";
+        value += "\"";
         escaped = false;
       } else if (quotedValue) {
-        quotedValue=false;
+        quotedValue = false;
       } else {
         quotedValue = true;
       }
-    } else if (str[i]=='\\'){
+    } else if (str[i] == '\\') {
       if (escaped) {
-        value << "\\";
+        value += "\\";
         escaped = false;
       } else {
         escaped = true;
       }
-    } else if (str[i]=='}') {
-      if(quotedValue){
-        value << "}";
+    } else if (str[i] == '}') {
+      if (quotedValue) {
+        value += "}";
       } else {
         write = true;
       }
     } else {
-      value<<str[i];
+      value += str[i];
     }
 
     if (write) {
-      strs.push_back(value.str());
-      value.str("");
+      strs.push_back(value);
       value.clear();
-      write=false;
+      write = false;
     }
   }
   return strs;


Woazboat:
Ideally the temporary (vector of) strings should not be created at all unless required by the consumer. In most cases the string does not need to be un-quoted and all that's done with the strings is convert them to integers. This could be done with e.g. a callback function that gets passed a string_view of the individual elements. That way we could avoid allocating and copying a std::string completely.

mmd-osm:
Thanks! I'm merging this right now.

I agree, when we're sure that the string doesn't use quoting, like a list of node ids, we might do a bit better with different approaches. As an example, absl::StrSplit (https://abseil.io/docs/cpp/guides/strings) uses a find + string_view.

I've already seen some good improvements with this change, so would leave it at that for the moment.

Woazboat:
I have a work-in-progress implementation of an array parsing iterator/view that can yield string_views of individual array elements on the fly, including for quoted values. Only backslash escaped values need a second un-escaping step to remove the escape characters since we can't modify the original string, but that can be done on-demand and only when required. It's about twice as fast compared to the `std::string` version here (for values without backslash escaping). It could be made even faster for arrays that are known to have no quoting/escaping since then all the checks for that could be optimized out.
https://github.com/Woazboat/openstreetmap-cgimap/commit/40c6c3e057808388202e6c6db1433e089e4f2170
https://quick-bench.com/q/urIpF-QeYLvVQTTcJ_JhcG_D2oM

mmd-osm:
std::function has some overhead, maybe you could also try a template parameter for the lambda function:

```
template <class Functor>
void psql_array_to_vector_callback(std::string_view str, Functor callback) {
```

Woazboat:
Yeah I know, it was originally just a function pointer for that reason since I was just testing it and didn't want to make it into a template at that point. I'm not going to work on the callback version any further though since the iterator/view is better all around, so it's just in there for comparison.

Woazboat:
It's about the same as the view version as a template https://quick-bench.com/q/G14I2eA16AIg5epd7wnEr_7dM1w

Woazboat:
> I have a work-in-progress implementation of an array parsing iterator/view that can yield string_views of individual array elements on the fly, including for quoted values. Only backslash escaped values need a second un-escaping step to remove the escape characters since we can't modify the original string, but that can be done on-demand and only when required. It's about twice as fast compared to the `std::string` version here (for values without backslash escaping). It could be made even faster for arrays that are known to have no quoting/escaping since then all the checks for that could be optimized out. [Woazboat@40c6c3e](https://github.com/Woazboat/openstreetmap-cgimap/commit/40c6c3e057808388202e6c6db1433e089e4f2170) https://quick-bench.com/q/urIpF-QeYLvVQTTcJ_JhcG_D2oM

I've now added the optimizations for values that are known to be non-null and/or do not need to be escaped. There's some nice speedups that can be gained here.
https://github.com/Woazboat/openstreetmap-cgimap/tree/psql-array-iterator
https://quick-bench.com/q/PoLf2HSTa0y5WsiMYrhq-Z_TpF4

mmd-osm:
So in places like https://github.com/zerebubuth/openstreetmap-cgimap/blob/master/src/backend/apidb/common_pgsql_selection.cpp#L158-L175, where we're creating new comments instances from multiple psql_array_to_vector, we probably need to change processing a bit when using iterators. Did you have any particular pattern in mind how you want to iterate over multiple arrays?

Woazboat:
Instead of advancing the index counter we can simply advance the individual iterators themselves, that shouldn't be a problem. The loop condition needs to check if any iterator has reached its end and the size mismatch check needs to be done after the loop (i.e. simply check if any of the iterators isn't at its end and still has elements left). 
C++23 https://en.cppreference.com/w/cpp/ranges/zip_view or https://en.cppreference.com/w/cpp/ranges/zip_transform_view would make that even easier, but unfortunately we can't use that yet.

mmd-osm:
> I've now added the optimizations for values that are known to be non-null and/or do not need to be escaped. There's some nice speedups that can be gained here.
https://github.com/Woazboat/openstreetmap-cgimap/tree/psql-array-iterator

I've started to move a few obvious bits of this branch to https://github.com/mmd-osm/openstreetmap-cgimap/tree/patch/cn_array_vector so that these changes don't get lost.

In some of my  /map.json testcases, another 5% performance improvement with the  iterator approach seems possible. Will need to do some more testing to get a better understanding about the potential improvements vs. the added code complexity.

Woazboat:
Thanks! I haven't completely forgotten about that branch, but it's unfortunately been on the backburner. I'll try and rebase it to get it up to date soon.

mmd-osm:
With the latest changes in master, the fixed sized string approach using std::array is in the 1% difference range (indistinguishable from measurement noise)…

Woazboat:
The main difference that I can see between the fixed array approach and the current method of filling the string char by char is that it could potentially avoid reallocations for large strings by constructing it all at once. I suspect that doesn't happen often enough to make a noticeable difference. 
It performs an additional copy to the temporary array before copying the full array to the internal string buffer while the direct append to the string can fill the internal string buffer directly. In theory this could actually make it slightly slower for short strings that use small string optimization and do not require heap (re-)allocation.

mmd-osm:
By the way, compared to v0.9.0, my map.json testcase is now about 33% faster when using gzip compression, and about 15% for identity (uncompressed). Brotli is a bit faster than gzip, and has even better compression.

psql_array_to_vector shows up in the 2-3% range with additional 2-3% for string / vector construction, which is triggered by psql_array_to_vector. I think it will be somewhat challenging to achieve further significant speed ups.

Woazboat:
Nice :+1: 
How are you testing the performance exactly? I've been thinking about adding some sort of benchmark that could be used as a reference to see the impact of changes.

mmd-osm:
Usually, I download some city level data from Overpass API (without the metadata), then use `osmium renumber -s -1 input.osm -o output.osm` to generate negative osm ids suitable for JOSM upload. In JOSM I upload the data to my local instance with all validations turned off (they are super slow and irrelevant in this case). That's good enough to upload millions of changes. For much larger data volumes, it's usually better to use some bulk upload script. In this case it wasn't needed. To measure /map.json performance I was using the Apache Benchmark tool, and linux perf tools for all analysis work, as well as https://github.com/KDAB/hotspot for fancy flame graphs.

mmd-osm:
@Zverik : https://en.osm.town/@zverik/112375772496506669 -> why not use osmium renumber to fix ids (see previous comment). Works like charm.

PR #354, opened 2023-12-20
OAuth 1.0a removal

mmd-osm:
This pull request completely removes OAuth 1.0a, including all test cases, config options, manual page details, etc.
Redaction related unit tests which were based on OAuth 1.0a before are now using OAuth 2.0.

As a side effect of this PR, we're reducing the number of db connections from 3 to 2 per process.

Maybe create a new release 0.8.11 before merging this PR.

**Don't merge right away, since it would impact some unit tests on master.apis.dev.**


Zverik:
Please set a generous deadline for this. There are too many tools on 1.0a yet. Migrating mine would take a couple months.

mmd-osm:
You need to talk to operations when this will be deployed. We’re only removing code here. 😎

mmd-osm:
We will wait a bit more for OrganicMaps to merge their OAuth 1.0a -> OAuth 2.0 PR. Target date is after their April release, so probably within in the next couple of weeks.

Otherwise merging this PR will break their unit tests on master.apis.dev.openstreetmap.org.

mmd-osm:
https://github.com/organicmaps/organicmaps/pull/7333 has been merged in the meantime -> "waiting downstream" label removed.

Issue #363, opened 2024-02-29
Last digit missing in bbox coordinate of changeset JSON call


nrenner:
The `minlon` and `maxlon` properties in the JSON response of the metadata API call for [changeset 146899741](https://www.openstreetmap.org/changeset/146899741) ([node 11572546705 v1](https://www.openstreetmap.org/node/11572546705/history/1)) are missing the last digit, the XML response has it:

`Actual  `: `-117.067166 ` - https://www.openstreetmap.org/api/0.6/changeset/146899741.json
`Expected`: `-117.0671669` - https://www.openstreetmap.org/api/0.6/changeset/146899741

This was noticed because the adiff query with that bbox in OSMCha fails to load the node, as [reported](https://osmus.slack.com/archives/C04Q8R2G17E/p1706907571545059) on Slack.

Woazboat:
Posting the full response data here for future reference:

```json
{"version":"0.6","generator":"CGImap 0.8.10 (2938081 spike-07.openstreetmap.org)","copyright":"OpenStreetMap and contributors","attribution":"http://www.openstreetmap.org/copyright","license":"http://opendatacommons.org/licenses/odbl/1-0/","elements":[{"type":"changeset","id":146899741,"created_at":"2024-01-31T11:17:34Z","closed_at":"2024-01-31T11:17:36Z","open":false,"user":"newportequipmentwa1","uid":20953528,"minlat":48.1730392,"minlon":-117.067166,"maxlat":48.1730392,"maxlon":-117.067166,"comments_count":0,"changes_count":1,"tags":{"changesets_count":"1","comment":"Excavating Contractor","created_by":"iD 2.27.3","host":"https://www.openstreetmap.org/edit","imagery_used":"Bing Maps Aerial","locale":"en-US"}}]}
```

```xml
<osm version="0.6" generator="CGImap 0.8.10 (2351905 spike-07.openstreetmap.org)" copyright="OpenStreetMap and contributors" attribution="http://www.openstreetmap.org/copyright" license="http://opendatacommons.org/licenses/odbl/1-0/">
<changeset id="146899741" created_at="2024-01-31T11:17:34Z" closed_at="2024-01-31T11:17:36Z" open="false" user="newportequipmentwa1" uid="20953528" min_lat="48.1730392" min_lon="-117.0671669" max_lat="48.1730392" max_lon="-117.0671669" comments_count="0" changes_count="1">
<tag k="changesets_count" v="1"/>
[...]

PR #364, opened 2024-02-29
Fix json output incorrectly truncating coordinates 

Woazboat:
Fixes https://github.com/zerebubuth/openstreetmap-cgimap/issues/363

The json formatter used a 12 byte buffer to format floating point values. This is not enough to hold the full string including the final null terminator byte for negative values with three digits before the decimal point and seven digits precision. The output of `snprintf()` was not checked, so if there is a digit which does not fit into the buffer it will be silently dropped.

E.g.  `-117.0671669` has 12 digits/bytes including the minus sign character. However, the terminating null character also requires 1 byte so the output string gets silently truncated to `-117.067166`.

This means that all coordinates with longitudes from -100 to -180, which is mostly the western USA and Canada, will be shifted by up to ~11cm at the equator (~8 cm in San Francisco) if downloaded via the json format. 
https://wiki.openstreetmap.org/wiki/Precision_of_coordinates#Precision_of_longitudes
longitude_-180_to_-100

### DIFF
--- a/src/json_writer.cpp
+++ b/src/json_writer.cpp
@@ -12,6 +12,9 @@
 #include <cstdio>
 #include <cstring>
 #include <utility>
+#include <array>
+#include <fmt/core.h>
+#include <fmt/compile.h>
 
 #include "cgimap/json_writer.hpp"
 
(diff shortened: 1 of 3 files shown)


mmd-osm:
Thanks a lot for the quick fix!

PR #369, opened 2024-03-18
Create separate cmake files for contrib libs

Woazboat:
Move contrib libs cmake definitions out of main cgimap cmake files into separate CMakeLists.txt files per library


mmd-osm:
Thanks, that's a a good idea!

PR #375, opened 2024-03-21
Added support for brotli response body compression

### LINKED ISSUE #374: Support for brotli compression
When compression was originally added to CGImap about 15 years ago, Brotli wasn't a thing yet. We could check, if we want it to add now. 

I started a bit of prototyping (no test cases yet): https://github.com/mmd-osm/openstreetmap-cgimap/tree/feature/brotli

mmd-osm:
* Brotli enabled by default in CMakeLists (is this a good idea?)
* Adds a new package dependency: libbrotli-dev
* Compression level 5 seems to be a good trade-off between compression rate and processing time.
* Brotli is the preferred option for `Accept-Encoding: gzip, compress, br` http header, like on the Rails port/mod_brotli on Apache.
* Changeset upload decompression is not in scope, maybe we could add this later.

---
Implementation details:

* cmake/FindBrotli.cmake was copied over from https://github.com/yhirose/cpp-httplib/blob/master/cmake/FindBrotli.cmake. It seemed like one of the more complete implementations that I could find out there.
* Brotli API calls in src/brotli.cpp were based on the implementation in https://github.com/yhirose/cpp-httplib/blob/master/httplib.h 

---
Closes #374

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,7 @@ endif()
 ###############
 option(ENABLE_APIDB "Enable APIDB backend, as used by the OSM servers" ON)
 option(ENABLE_YAJL "Enable JSON output with the YAJL library" ON)
+option(ENABLE_BROTLI "Enable Brotli library" ON)
 option(ENABLE_FMT_HEADER "Enable FMT header only mode" ON)
 option(ENABLE_COVERAGE "Compile with coverage info collection" OFF)
 option(ENABLE_PROFILING "Compile with profiling" OFF)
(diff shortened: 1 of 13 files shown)


mmd-osm:
@tomhughes : before merging this PR, it would be good to know, if `libbrotli-dev` is available on faffy. Although I don't see it listed in chef, the relevant files such as filebrotli/decode.h , libbrotlicommon.so.1 and libbrotlienc.so.1 might be pulled in by some other dependency already. TIA.

tomhughes:
It was actually installed but I've added it to chef anyway.

mmd-osm:
Thanks, I'll give it a try now. If there are some bugs during testing, we'll address them in another PR.

PR #381, opened 2024-04-01
Refactor cmake install and package definitions

Woazboat:
* Use implicit default values where possible
* Simplify cmake install targets via GNUInstallDirs module
  * Installation directories for binaries, libaries, etc... are set automatically according to their target type
* Define components for install/package targets
  * Allows building separate packages for binaries and development headers (`*-dev` package)
* Rename included modified libxml++ to differentiate it from original libxml++

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -178,7 +178,7 @@ if(ENABLE_INSTALL)
     set(CGIMAP_INSTALL_INCLUDEDIR ${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME})
 
     if(BUILD_SHARED_LIBS)
-        set(CGIMAP_LIBS cgimap_core cgimap_fcgi cgimap_apidb libxml++)
+        set(CGIMAP_LIBS cgimap_core cgimap_fcgi cgimap_apidb cgimap_libxml++)
     endif()
 
     if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.23.0)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -185,7 +185,8 @@ if(ENABLE_INSTALL)
         set(CGIMAP_HEADER_FILE_SET_INSTALL_ARTIFACTS
             FILE_SET HEADERS
                 COMPONENT Development
-                DESTINATION ${CGIMAP_INSTALL_INCLUDEDIR})
+                DESTINATION ${CGIMAP_INSTALL_INCLUDEDIR}
+                EXCLUDE_FROM_ALL)
     endif()
 
     install(TARGETS openstreetmap_cgimap ${CGIMAP_LIBS}
@@ -196,9 +197,11 @@ if(ENABLE_INSTALL)
         PUBLIC_HEADER
             COMPONENT Development
             DESTINATION ${CGIMAP_INSTALL_INCLUDEDIR}
+            EXCLUDE_FROM_ALL
         PRIVATE_HEADER
             COMPONENT Development
             DESTINATION ${CGIMAP_INSTALL_INCLUDEDIR}
+            EXCLUDE_FROM_ALL
         ${CGIMAP_HEADER_FILE_SET_INSTALL_ARTIFACTS})
 
     install(FILES openstreetmap-cgimap.1 TYPE MAN
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -167,20 +167,36 @@ endif()
 # Install
 #############################################################
 
-if (ENABLE_INSTALL)
-  install(TARGETS openstreetmap_cgimap
-          PERMISSIONS OWNER_EXECUTE OWNER_WRITE OWNER_READ GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE
-          DESTINATION "bin")
+if(ENABLE_INSTALL)
+    include(GNUInstallDirs)
 
-  install(FILES openstreetmap-cgimap.1 DESTINATION share/man/man1)
+    set(CMAKE_INSTALL_DEFAULT_DIRECTORY_PERMISSIONS
+        OWNER_READ OWNER_WRITE OWNER_EXECUTE
+        GROUP_READ GROUP_EXECUTE
+        WORLD_READ WORLD_EXECUTE)
 
-  install(FILES README DESTINATION share/doc/openstreetmap-cgimap)
+    set(CGIMAP_INSTALL_INCLUDEDIR ${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME})
 
-  if (BUILD_SHARED_LIBS)
-    install(TARGETS cgimap_core DESTINATION lib)
-    install(TARGETS cgimap_fcgi DESTINATION lib)
-    install(TARGETS cgimap_apidb DESTINATION lib)
-  endif()
+    if(BUILD_SHARED_LIBS)
+        set(CGIMAP_LIBS cgimap_core cgimap_fcgi cgimap_apidb libxml++)
+    endif()
+
+    if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.23.0)
+        set(CGIMAP_HEADER_FILE_SET_INSTALL_ARTIFACTS
+            FILE_SET HEADERS
+                DESTINATION ${CGIMAP_INSTALL_INCLUDEDIR})
+    endif()
+
+    install(TARGETS openstreetmap_cgimap ${CGIMAP_LIBS}
+        PUBLIC_HEADER
+            DESTINATION ${CGIMAP_INSTALL_INCLUDEDIR}
+        PRIVATE_HEADER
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -176,21 +177,48 @@ endif()
...
-  install(TARGETS openstreetmap_cgimap
-          PERMISSIONS OWNER_EXECUTE OWNER_WRITE OWNER_READ GROUP_READ GROUP_EXECUTE WORLD_READ WORLD_EXECUTE
-          DESTINATION "bin")
+    include(GNUInstallDirs)
 
-  install(FILES openstreetmap-cgimap.1 DESTINATION share/man/man1)
+    set(CMAKE_INSTALL_DEFAULT_DIRECTORY_PERMISSIONS
+        OWNER_READ OWNER_WRITE OWNER_EXECUTE
+        GROUP_READ GROUP_EXECUTE
+        WORLD_READ WORLD_EXECUTE)
```
Woazboat, line 185: I coped this over from the existing code, but is this really required? The correct permissions should already be set automatically.
(I changed `CPACK_INSTALL_DEFAULT_DIRECTORY_PERMISSIONS` to `CMAKE_INSTALL_DEFAULT_DIRECTORY_PERMISSIONS` which is also used by cpack by default).

mmd-osm:
OSM operations wants to go ahead with Ubuntu 24.04 for the time being and only move to Debian at a later point in time. This in turn means that cpack has only low prio now, since we would continue using Ubuntu Launchpad for packaging and shipping PPAs to OSM production.

tomhughes:
That was intended more as a general comment on our strategy, that in the short term we are likely to upgrade machines to 22.04 simply because moving them to Debian requires a reinstall which is significantly more complicated while new hardware would get Debian.

That said as the web frontends are duplicated we may well want to move them to Debian at some point because we can do that in a staged manner without disrupting service.

mmd-osm:
So there's some update in https://osmfoundation.org/w/index.php?title=Operations/Minutes/2024-06-27&oldid=12823#%22OSM_PPA%22_work_approval

Minutes say "CGImap is blocking us moving with the six front ends.". I don't really know what this is about, maybe @Firefishy knows more?

Firefishy:
> Minutes say "CGImap is blocking us moving with the six front ends.". I don't really know what this is about, maybe @Firefishy knows more?

We'd like a PPA like deb repo for Debian 12 with our own builds (with potential extra patches). A replica of https://launchpad.net/~osmadmins/+archive/ubuntu/ppa

mmd-osm:
Thank you for the quick reply, this makes sense to me. Where exactly is CGImap blocking you atm?

Woazboat:
Afaik an apt repo is more or less just a bunch of files on a http server, so that could potentially be selfhosted?

tomhughes:
It is from the point of view of serving it, yes. The complicated part it seems is turning a bunch of debs into a working file tree to serve.

I'm used to RPM where you can chuck a bunch of rpms in a directory, run createrepo to create the metadata and then just serve that up but the Debian wiki page on doing the equivalent is about 3 miles long with about 30 different ways of doing it none of which seem to be anything like that simple.

Woazboat:
I just played around with https://www.aptly.info/ a bit and it seems fairly easy to use. [^1]

```sh
aptly repo create -distribution=stable -component=main -comment="Debian repository for openstreetmap-cgimap" openstreetmap-cgimap
aptly repo add openstreetmap-cgimap /path/to/deb/files/
aptly publish repo openstreetmap-cgimap
```

[^1]: Listed in https://wiki.debian.org/DebianRepository/Setup

mmd-osm:
This is a bit funny. I tried out aptly pretty much at the same time, and also thought it could be a good starting point for further evaluation. Creating  ".deb" files for Debian 12 should be feasible, since we're anyway running test cases + build binaries in a Debian Bookworm docker image.

Woazboat:
All you need to do to create a .deb file is run `make package`. I actually have an additional patch that builds on this pull request that uses the created .deb packages to install cgimap in the docker images. That also automatically takes care of installing the required runtime dependencies without having to list and update them all manually in the dockerfile.

https://github.com/zerebubuth/openstreetmap-cgimap/commit/d499553211747d8df75cd6647d04769e03aa0c89#diff-a63453f375af09161776e2080bcca906cd03655f476bd2c024dbaee0dfe9f94f

tomhughes:
Thanks for the suggestion - we now have an aptly based repo live at https://apt.openstreetmap.org/ which has replaced our PPA in production.

CMakeLists.txt
```diff
@@ -167,20 +167,36 @@ endif()
...
-  if (BUILD_SHARED_LIBS)
-    install(TARGETS cgimap_core DESTINATION lib)
-    install(TARGETS cgimap_fcgi DESTINATION lib)
-    install(TARGETS cgimap_apidb DESTINATION lib)
-  endif()
+    if(BUILD_SHARED_LIBS)
+        set(CGIMAP_LIBS cgimap_core cgimap_fcgi cgimap_apidb libxml++)
+    endif()
+
+    if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.23.0)
```
mmd-osm, line 184: I tested static and shared builds, but my local cmake version seems a bit too old to test the header files. What is the expected behavior here? Will be always include header files in the .deb package?

tomhughes:
Moving the frontends to Debian 12 or Ubuntu 24.04 is getting a bit more urgent so do you have any idea when you might be in a position to provide packages?

mmd-osm:
Re-publishing the current 0.9.3 release for Ubuntu 24.04 should be easy and shouldn't take too much time (again on Ubuntu launchpad). I still haven't figured out all details for the new apt.osm.org. Falling back to Ubuntu launchpad seems like a safe bet at the moment. Would that be ok?

tomhughes:
We don't have any support for publishing to apt.osm.org yet so for now just build the packages wherever and I can load them manually.

We'd probably prefer just to jump to Debian 12 but I realise that's probably a bigger ask than Ubuntu.

mmd-osm:
Ok, good. It seems getting the build up and running for Debian 12 should have priority over anything Ubuntu related.

In https://github.com/mmd-osm/openstreetmap-cgimap/tree/test/381 I have tested this PR with some more changes to generate and deploy a .deb package for Debian 12, with unit tests enabled.

The resulting docker image has the .deb file stored underneath /app_deb. This directory could be copied to the host machine, and then uploaded to apt.osm.org. Unfortunately, the package signing topic is not covered at the moment, and needs to be fixed urgently.

It would be good to automate more of this as part of GitHub actions. As we have discussed in some other issue, my collaborator role is way too limited to do the proper config. We either need zerebubuth for that, or move the whole repo elsewhere.

mmd-osm:
Storing .deb packages as build artifacts seems to work in general: https://github.com/mmd-osm/openstreetmap-cgimap/actions/runs/10390754008

CMakeLists.txt
Woazboat: The header files will only be included in the package if 
1) the library targets containing them are being installed (only in a shared build here) 
and 
2) the `Development` component is enabled (enabled by default)

There was actually a name mismatch in the install command file set (s/HEADER/HEADERS/) that I fixed now. That's probably why you didn't see the header files.

I changed it now so that headers won't be included by default in the standard all-in-one install/package, only when a component based installation is performed (`CPACK_DEB_COMPONENT_INSTALL` is set to `YES`). With per-component packages you'll a different .deb file for each component that you can install individually, so installing the `*-dev` header package is optional.

mmd-osm: Thank you for looking into this issue! Somehow, I still didn't make to generate multiple package

```
cmake .. -DBUILD_SHARED_LIBS=YES -DCPACK_DEB_COMPONENT_INSTALL=YES`
```

I'm pretty sure this is due to my cmake version 3.22.1. Now that the headers are not included by default, I'm good to merge this PR right now. I will add some further changes from test/381 branch to store Debian packages as build artifact to Debian 12 at least. Operations is waiting for fresh deb packages...

CMakeLists.txt
```diff
@@ -7,14 +7,15 @@ set(CMAKE_DISABLE_SOURCE_CHANGES ON)
...
-# CGImap project
+# openstreetmap-cgimap project
 ################
 string(TIMESTAMP CURRENT_TIMESTAMP %y%m%d%H%M UTC)
 
-project(CGImap LANGUAGES CXX
-               VERSION "0.10.0.${CURRENT_TIMESTAMP}"
-               DESCRIPTION "CGImap is a C++ implementation of some parts of the OpenStreetMap API as a FastCGI process."
-               HOMEPAGE_URL "https://github.com/zerebubuth/openstreetmap-cgimap")
+project(openstreetmap-cgimap
```
mmd-osm, line 14: As a side note, this change is also visible in the generator header attribute returned by the endpoints:

generator="openstreetmap-cgimap 2.0.0 ..."

vs.

 generator="CGImap 0.9.3 ..."

PR #382, opened 2024-04-01
Integrate clang-tidy in cmake build

Woazboat:
Add option to automatically run clang-tidy during compilation. Currently disabled by default due to the number of warnings it produces. The enabled checks can be configured in `.clang-tidy`.

Also adds some more options for other linters that are directly supported by cmake.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,3 @@
+---
+Checks: '-*,cppcoreguidelines-*,bugprone-*,performance*,modernize-*,-modernize-use-trailing-return-type'
+...
--- a/.iwyu_mappings.imp
+++ b/.iwyu_mappings.imp
@@ -1,0 +1,10 @@
+[ 
+  { include: ["@<libmemcached-.*/.*>", "public", "<libmemcached/memcached.h>", "public"] },
+  { include: ["@<boost/algorithm/string/.*>", "public", "<boost/algorithm/string.hpp>", "public"] },
+  { include: ["@<boost/program_options/.*>", "public", "<boost/program_options.hpp>", "public"] },
+  { include: ["@<boost/lexical_cast/.*>", "public", "<boost/lexical_cast.hpp>", "public"] },
+  { include: ["@<pqxx/.*\\.h.*>", "private", "<pqxx/pqxx>", "public"] },
+  { include: ["<cryptopp/config_int.h>", "private", "<cryptopp/config.h>", "public"] },
+  { include: ["<bits/chrono.h>", "private", "<chrono>", "public"] },
+  { include: ["<bits/local_lim.h>", "private", "<limits.h>", "public"] },
+] 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,6 +53,7 @@ endif()
 
 include(CTest)
 include(AutotoolsCompatibilityDefinitions)
+include(BuildLint)
 
 #########################
 # common compiler options
@@ -145,17 +146,12 @@ target_link_libraries(openstreetmap_cgimap
     PQXX::PQXX)
 
 
-
 #############################################################
 # Optional "clang-tidy" target
 #############################################################
 
-find_program(CLANG_TIDY
-             NAMES clang-tidy clang-tidy-16 clang-tidy-15 clang-tidy-14 clang-tidy-13 clang-tidy-12 clang-tidy-11)
-
-if (CLANG_TIDY)
-    message(STATUS "Looking for clang-tidy - found ${CLANG_TIDY}")
-
+# BuildLint module already checks for clang-tidy
+if(CLANG_TIDY)
     file(GLOB CT_CHECK_FILES src/*.cpp src/api06/*.cpp src/api06/changeset_upload/*.cpp src/api07/*.cpp
                              src/backend/apidb/*.cpp src/backend/apidb/changeset_upload/*.cpp
                              test/*.cpp test/*.hpp)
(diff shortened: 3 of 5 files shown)


mmd-osm:
Thanks for this. So far, I'm running at least cppcheck manually, from time to time also clang-tidy. It makes sense to have it automated, so let's give it a try... will update this comment as I'm testing different bits.

* clang-tidy (-14):
   + it would be good to exclude the contrib/ directory, e.g .by adding a contrib/.clang-tidy file with `Checks: '-*'`
  + i'm getting lots of failures like `no template named 'optional' in namespace 'std'; did you mean 'boost::optional'? [clang-diagnostic-error]` that look like clang-tidy isn't aware that we're using C++17.

Woazboat:
Explicitly set the c++ standard version for clang-tidy now. Do we want `c++17` or `gnu++17`?

I haven't found cpplint to be that useful either with all checks enabled. I've looked up how to configure it now and removed some of the noisy checks. Formatting changes would probably best be done via a separate clang-format pass.

I also disabled the build-time linter checks for the contrib libraries.

Cmake is currently configured to go ahead anyway even if a linter is not installed. A 'not-found' message is printed during configuration. That could be changed to abort the build if a program is missing or changed to a stronger warning.

mmd-osm:
Now with the latest changes in place, I've tested all linters again. Results look much better now. I'm was mostly struggling to get iwyu working as expected, the other three linters are good to go.

Boost was triggering a few cycles in include-mapping, which I couldn't resolve. Maybe my iwyu version is too old?

```
[ 13%] Building CXX object src/CMakeFiles/cgimap_core.dir/choose_formatter.cpp.o
Error running '/usr/bin/iwyu': Cycle in include-mapping:
  <boost/preprocessor/iteration/detail/iter/forward1.hpp> ->
  <boost/spirit/home/support/detail/as_variant.hpp> ->
  <boost/preprocessor/iteration/detail/iter/forward1.hpp> ->
  <boost/preprocessor/iteration/detail/iter/forward1.hpp>
/build/iwyu-5ZxwEx/iwyu-8.17/iwyu_include_picker.cc:1081: Assertion failed: Cycle in include-mapping
```

```
[ 46%] Building CXX object src/CMakeFiles/cgimap_core.dir/api06/handler_utils.cpp.o
Error running '/usr/bin/iwyu': Cycle in include-mapping:
  <boost/spirit/home/support/detail/as_variant.hpp> ->
  <boost/preprocessor/iteration/detail/iter/forward1.hpp> ->
  <boost/preprocessor/iteration/detail/iter/forward1.hpp>
/build/iwyu-5ZxwEx/iwyu-8.17/iwyu_include_picker.cc:1081: Assertion failed: Cycle in include-mapping
```

iwyu was a bit too eager and resolved the following include:

```
 #include <libmemcached/memcached.h>  // lines 11-11
``` 

to this list of includes:
 
```
#include <libmemcached-1.0/behavior.h>              // for memcached_behavio...
#include <libmemcached-1.0/deprecated_types.h>      // for memcached_return
#include <libmemcached-1.0/get.h>                   // for memcached_get
#include <libmemcached-1.0/memcached.h>             // for memcached_create
#include <libmemcached-1.0/parse.h>                 // for memcached_servers...
#include <libmemcached-1.0/server_list.h>           // for memcached_server_...
#include <libmemcached-1.0/storage.h>               // for memcached_add
#include <libmemcached-1.0/types/behavior.h>        // for MEMCACHED_BEHAVIO...
#include <libmemcached-1.0/types/return.h>          // for MEMCACHED_NOTSTORED
```

It would be good if we could exlucde libmemcached/* from the analysis.  `--keep=<glob>` might be good for this?

In a similar way, I'd like to avoid introducing forward declarations. Ideally, `--no_fwd_decls` should do the job. Unfortunately, I didn't have much success in adding both options to BuildLint.cmake.

```
include/cgimap/api06/changeset_upload/osmchange_handler.hpp should add these lines:
namespace api06 { class Node; }
namespace api06 { class Node_Updater; }
namespace api06 { class OSMObject; }
namespace api06 { class Relation; }
namespace api06 { class Relation_Updater; }
namespace api06 { class Way; }
namespace api06 { class Way_Updater; }
```


> Do we want c++17 or gnu++17?
I think c++17 should be ok. I don't think there are differences which would be relevant for us, but I haven't really tested it.

> Cmake is currently configured to go ahead anyway even if a linter is not installed. A 'not-found' message is printed during configuration. That could be changed to abort the build if a program is missing or changed to a stronger warning.

Yes, I think it's ok to have the message during configuration only. I missed this warning since I was using ccmake and screen contents had already disappeared before I could read all warnings. I do remember that iwyu was shown with some "NOTFOUND" -like filename.

Woazboat:
I've added some parameters and header file mappings for iwyu that should make it a bit more useful. I haven't found a way to get rid of the include cycle errors unfortunately.
The output of iwyu is still rather noisy and should be taken with a big grain of salt. The cmake integration should best be used as a convenient way to run it for occasional manual spot checks and not as a permanent feature in every build.

mmd-osm:
I tried to add a few boost*.imp files from https://github.com/include-what-you-use/include-what-you-use as additional mapping files, but that also didn't help. 

As an alternative some people seem to be using iwyu_tool for single file analysis:

```
cmake .. -D CMAKE_EXPORT_COMPILE_COMMANDS=ON
make -j
iwyu_tool  -p build src/fcgi_request.cpp -- -std=c++17 -Xiwyu --cxx17ns -Xiwyu --no_fwd_decls -Xiwyu --no_comments  -Xiwyu --mapping_file=`pwd`/.iwyu_mappings.imp
```

mmd-osm:
Thanks again for this PR. I'm merging this now!

PR #386, opened 2024-04-10
Use std::forward for forwarding references instead of std::move

Woazboat:


mmd-osm:
By the way, the generated code looked exactly the same...

Woazboat:
In this case it's good that it did because otherwise the old code would have had a bug. `std::move()` forces the reference to be a temporary/rvalue reference and could cause an incorrect move and invalidation of the original even if it was actually an lvalue that needs to be copied. `std::forward` prevents that from happening by preserving the rvalue/lvalue.

PR #392, opened 2024-04-14
Add section about testing API calls with `cgi-fcgi` tool to readme

Woazboat:
Added a section about [cgi-fcgi](https://manpages.debian.org/testing/libfcgi-bin/cgi-fcgi.1.en.html) to the readme as it can be quite useful for quick testing of API calls.
One example I included uses basic auth which will be removed soon, but I did not yet have the time to provide a full example including how to get tokens for oauth2.

### DIFF
--- a/README
+++ b/README
@@ -187,6 +187,8 @@ CGImap to avoid any possibility of data corruption.
 
 ## Testing
 
+### Test suite
+
 To run the test suite you will need additional packages installed:
 
     sudo apt-get install postgresql-all postgresql-common postgresql-server-dev-all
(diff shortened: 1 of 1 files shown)


mmd-osm:
For some reason, we're already using cgi-fcgi in this test script: https://github.com/zerebubuth/openstreetmap-cgimap/blob/master/test/test-single.sh - though I could never really figure out what it's supposed to do. I even thought of getting rid of it.

I'm not really a big fan of cgi-fcgi I have to admit, since it's very cumbersome to use and not suitable beyond a single API call (rules out testing in iD or JOSM). I use lighttpd for local testing, which I can run as normal user, kill / restart whenever needed. Either I let lighttpd start up openstreetmap-cgimap, or I start cgimap in daemon mode, and have lighttpd point at the FastCGI port.

Is there anything we could do to improve the documentation for lighttpd instead?

Woazboat:
I don't see them as mutually exclusive. A full http server like lighttpd will of course be better and more useful once it's set up correctly. However, the initial setup is more complicated and requires more steps and more configuration where things can go wrong. 
`cgi-fgci` provides a 'dumb' way of testing API calls that doesn't have many moving parts that can fail and doesn't require running another different server in the background. That can be quite useful for simple checks or even just for debugging issues with e.g. the lighttpd setup.

Re. that test script: I'm also not entirely sure how it's supposed to be used. It looks like a wrapper for running other test scripts that do the actual evaluation, but I don't see anything else in the repo that could be used in combination with it. It can probably be removed.

mmd-osm:
Regarding the OAuth2 token generation, it might be a good idea to either create a rake task on the openstreetmap-website project that creates some sample users, including tokens. Creating users, assigning roles, registering OAuth 2 apps, and generating tokens could all be automated by something similar to:

```ruby
def oauth_token(application_id, user_id)
  application = Doorkeeper.config.application_model.find_by(:uid => application_id)

  Doorkeeper.config.access_token_model.find_or_create_for(
    :application => application,
    :resource_owner => user_id,
    :scopes => application.scopes
  )
end

password = "openstreetmap"
admin = User.create(:display_name => "admin", :email => "admin2@demo.cc", 
                    :pass_crypt => password, :pass_crypt_confirmation => password,
                    :tou_agreed => Time.now.utc, :terms_seen => true,
                    :terms_agreed => Time.now.utc, :email_valid => true)
admin.activate!
admin.roles.create(:role => "administrator", :granter_id => admin.id)

testuser = User.create(:display_name => "testuser", :email => "testuser@demo.cc", 
                    :pass_crypt => password, :pass_crypt_confirmation => password,
                    :tou_agreed => Time.now.utc, :terms_seen => true,
                    :terms_agreed => Time.now.utc, :email_valid => true)
testuser.activate!

ideditor = Doorkeeper::Application.create(name: "Local iD", owner_type: "User", owner_id: User.find_by(:display_name => "testuser").id, 
                                          redirect_uri: "http://localhost:3000",
                                          scopes: ["read_prefs write_prefs write_api read_gpx write_gpx write_notes"])

puts "Token: #{ oauth_token(ideditor.uid, User.find_by(:display_name => "testuser").id).token }"

```

gstrauss:
If it helps, lighttpd can be run in a one-shot mode where the requests are fed into stdin.  Debian autopkgtests for lighttpd uses this.

lighttpd's own test infrastructure runs lighttpd on an ephemeral port so that tests can be run regardless of what other daemons are running on the server running the tests.

mmd-osm:
@gstrauss : thanks a lot for the excellent proposal, I wasn't even aware of this option. I found one of your posts [here](https://github.com/zyw-200/FirmAFL/issues/32) with a bit more context. Previous example could be rewritten as:

```
echo -n 'PUT /api/0.6/changeset/create HTTP/1.1
Host: foo
Content-Type: text/plain
Authorization: Bearer u_i2......
Content-Length: 72

<osm><changeset><tag k="created_by" v="cgimap_test" /></changeset></osm>' | unix2dos | lighttpd -1 -f  configfile
```

(I included unix2dos to avoid adding all these \r\n manually)

Woazboat:
That seems like a better option for simple one-shot tests compared to cgi-fcgi

PR #398, opened 2024-04-21
Fix libxml++ contrib library compiler options

Woazboat:
Linking the libxml++ contrib library with cgimap_common_compiler_options with INTERFACE scope is a bug as that only applies to targets using the library but not the library itself.
(Copy paste error from catch2 which is a header only interface library and can therefore only link to other libraries with INTERFACE scope)

https://github.com/zerebubuth/openstreetmap-cgimap/commit/0901120a963542b457ceb1f45890e18de4b4be88#diff-f3c6529ba39f2f8fe71ba721bdb3eccec46bd4ba024980398b12c7e7cd3f2672R5

### DIFF
--- a/contrib/CMakeLists.txt
+++ b/contrib/CMakeLists.txt
@@ -5,4 +5,4 @@ add_subdirectory(catch2)
 add_subdirectory(libxml++)
 
 target_link_libraries(catch2 INTERFACE cgimap_common_compiler_options)
-target_link_libraries(libxml++ INTERFACE cgimap_common_compiler_options)
+target_link_libraries(libxml++ PUBLIC cgimap_common_compiler_options)


mmd-osm:
I'm a bit curious where this would show up today as an issue. Is it maybe that profiling and coverage would not work for libxml++, or would I notice some other issues?

Woazboat:
I don't think it currently has any effect other than profiling / coverage. C++17 is the default for >= gcc 11 anyway and the other defines `PACKAGE_NAME=...`, `PACKAGE_VERSION=...`, ... that are set aren't used in the libxml++ code. I noticed it because I was testing compilation with C++ 20 and that wasn't applied.

mmd-osm:
Yes, that makes sense, thanks for your feedback!

PR #407, opened 2024-05-25
JSON support for changeset upload

mmd-osm:
Back in https://github.com/openstreetmap/openstreetmap-website/issues/3501 we've talked about adding JSON support for a few remaining endpoints. Unfortunately, this project never took off. Why not give it another try now.

I found a streaming json parser based on yajl called [sjparser](https://gitlab.com/dhurum/sjparser), that has schema support. It offers a much saner interface than yajl alone, and emits on OSM element at a time. It also hides all that ugly state handling behind the scenes, unlike the XML parser we have in place today.

* Define new format specifications for JSON based messages:
   - /api/0.6/changeset/nnn/upload
       +  Input: OSMChange, based on OSM JSON format -> some examples in test_parse_osmchange_json_input.cpp
       +  Output: diffResult: TODO

There are a few other changeset related endpoints (create / update) that don't support JSON. This can be addressed in another PR.


mmd-osm:
Created an issue for sjparser: https://gitlab.com/dhurum/sjparser/-/issues/1

mmd-osm:
json_upload

PR #410, opened 2024-06-17
Add importer role

Woazboat:


mmd-osm:
Relevant upstream user role definition: https://github.com/openstreetmap/openstreetmap-website/blob/1a5adb2a4afba349edabce6ed84883df0b9a1879/app/models/user_role.rb#L26

mmd-osm:
Thanks, LGTM!

PR #411, opened 2024-06-17
Propagate request context to handlers

Woazboat:
Make context information about the request (user id, roles, ...) available in change handlers.
Unblocks future features like https://github.com/zerebubuth/openstreetmap-cgimap/pull/409#issuecomment-2171787640

### DIFF
--- a/include/cgimap/backend/apidb/changeset_upload/node_updater.hpp
+++ b/include/cgimap/backend/apidb/changeset_upload/node_updater.hpp
@@ -20,11 +20,14 @@
 #include <memory>
 #include <set>
 
+struct RequestContext;
+
 
 class ApiDB_Node_Updater : public api06::Node_Updater {
 
 public:
   ApiDB_Node_Updater(Transaction_Manager &_m,
+                     const RequestContext& req_ctx,
                      api06::OSMChange_Tracking &ct);
 
   ~ApiDB_Node_Updater() override = default;
@@ -108,6 +111,7 @@ private:
   void delete_current_node_tags(const std::vector<osm_nwr_id_t> &ids);
 
   Transaction_Manager &m;
+  const RequestContext& req_ctx;
   api06::OSMChange_Tracking &ct;
 
   std::vector<node_t> create_nodes;
--- a/include/cgimap/request_context.hpp
+++ b/include/cgimap/request_context.hpp
@@ -1,0 +1,37 @@
+/**
+ * SPDX-License-Identifier: GPL-2.0-only
+ *
+ * This file is part of openstreetmap-cgimap (https://github.com/zerebubuth/openstreetmap-cgimap/).
+ *
+ * Copyright (C) 2009-2024 by the CGImap developer community.
+ * For a full list of authors see the git log.
+ */
+
+#ifndef REQUEST_CONTEXT_HPP
+#define REQUEST_CONTEXT_HPP
+
+#include "cgimap/types.hpp"
+
+#include <optional>
+#include <set>
+#include <chrono>
+
+struct request;
+
+struct UserInfo
+{
+    osm_user_id_t id = {};
+    std::set<osm_user_role_t> user_roles = {};
+    bool allow_api_write = false;
+
+    bool has_role(osm_user_role_t role) const { return user_roles.count(role) > 0; }
+};
+
+struct RequestContext
+{
+    request& req;
+    std::chrono::system_clock::time_point start_time = {};
+    std::optional<UserInfo> user = {};
+};
+
+#endif /* REQUEST_CONTEXT_HPP */
--- a/test/test_apidb_backend_changeset_uploads.cpp
+++ b/test/test_apidb_backend_changeset_uploads.cpp
@@ -31,6 +31,7 @@
 #include "cgimap/process_request.hpp"
 #include "cgimap/output_buffer.hpp"
 #include "cgimap/zlib.hpp"
+#include "cgimap/request_context.hpp"
 
 #include "cgimap/api06/changeset_upload/osmchange_handler.hpp"
 #include "cgimap/api06/changeset_upload/osmchange_input_format.hpp"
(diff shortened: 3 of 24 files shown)


mmd-osm:
Thanks, this looks good even outside the context of #409. Will probably merge this PR later tonight.

include/cgimap/backend/apidb/changeset_upload/node_updater.hpp
```diff
@@ -20,11 +20,14 @@
...
 #include <set>
 
+struct RequestContext;
+
 
 class ApiDB_Node_Updater : public api06::Node_Updater {
 
 public:
   ApiDB_Node_Updater(Transaction_Manager &_m,
+                     const RequestContext& req_ctx,
```
mmd-osm, line 30: I've noticed that you're passing the new request context to {node,way,relation}_updater, where we previously didn't pass any of the information. Did you have some specific use cases in mind where this might come in handy? Not really a big issue, just wanted to understand the motivation behind this change.

mmd-osm:
As mentioned in https://github.com/zerebubuth/openstreetmap-cgimap/pull/409#issuecomment-2178096072, we're using the same database function approach as before with the changeset rate limit. For this reason, this pull request would no longer be a prerequisite for making progress on the topic.

I still find the general idea really useful, but would likely merge it only after we have released the next 0.9.x version to keep the amount of code to be backported as small as possible. I hope that's fine for now.

include/cgimap/backend/apidb/changeset_upload/node_updater.hpp
Woazboat: I don't think a simple bounding box size check will be enough to combat vandalism and we could certainly use more tools in our toolbox for that purpose. Providing the request context here where the uploaded objects are processed allows for checks and filters that take the actual map data into account (like e.g. a word filter for tags, region or object specific protections, etc...)
As I mentioned over in https://github.com/zerebubuth/openstreetmap-cgimap/pull/409#issuecomment-2171913919 I'm currently tinkering with a framework for checks like that which would require this information. [^1] It will certainly still take a while though until (/if) that is complete and I can carry these changes in my own branch anyway, so having these changes upstreamed is not really necessary at the moment if they aren't needed immediately.

[^1]: Although the biggest hurdle there is that we never actually have a complete look at the uploaded data since all changes are pushed to the database in chunks as soon as they are parsed.

mmd-osm: I also have some doubts that the bounding box checks will be sufficient in the end, or if it would even cause too much collateral damage. The approach I started looking into is to throw a list of created/modified/deleted object over the fence, and let some db function do the impact analysis, like what is total way length, how much displacement do we see. 

We could leverage PostGIS to do all sorts of geo analysis, and even include ML stuff through the plpython3u extension. These are all theoretical options, and it highly depends on what operations thinks is reasonable. 

Since we both see the to-be committed data, as well as the previous data on the database, chunked processing would not so much of an issue. I also want to keep the CGImap code clean and would try to keep complicated analysis logic outside the project as much as possible.

That's just my opinion. It's good to have different alternative ideas out there, and see what works best. Somehow we also need to consolidate all the requirements and then decide what data points would be needed in each case. This might help us to avoid too many "black swan" like situations in the future.

mmd-osm:
OAuth 1.0a / Basic auth retirement is now scheduled for July 1st [link](https://community.openstreetmap.org/t/oauth-1-0a-and-http-basic-auth-shutdown/108490/31). That means backporting will no longer be a concern in the near future.

include/cgimap/request_context.hpp
```diff
@@ -0,0 +1,37 @@
...
+    osm_user_id_t id = {};
+    std::set<osm_user_role_t> user_roles = {};
+    bool allow_api_write = false;
+
+    bool has_role(osm_user_role_t role) const { return user_roles.count(role) > 0; }
+};
+
+struct RequestContext
+{
+    request* req = nullptr;
```
mmd-osm, line 32: When checking the new signature for process_post_put_request, I was wondering why we're passing both request and RequestContent to that method. Then I realized that RequestContext.req is not being used at all. Is this intended as is? 

For the JSON upload I was looking to have access to the request headers inside changeset_upload_handler, which doesn't seem to be possible at the moment, since the instance variable is not being set.

Woazboat: I could swear I set that :thinking: 
Either it got lost somewhere during refactoring or it was just my imagination. 
I'll change the function to use just the RequestContext as well.

test/test_apidb_backend_changeset_uploads.cpp
```diff
@@ -161,9 +162,11 @@ TEST_CASE_METHOD( DatabaseTestsFixture, "test_single_nodes", "[changeset][upload
 
   SECTION("Create new node")
   {
+    test_request req{};
```
mmd-osm, line 165: I believe we could avoid the repeated  `test_request req{};  RequestContext ctx{req};` declarations here by moving these two lines to line 142 (right after TEST_CASE_METHOD.

During test execution, these "top level" declarations would be automatically re-executed for each SECTION, so that each SECTION would use a fresh ctx object.

In cases where sharing does not make sense (e.g. ctx2), I would leave them where they are today.

Woazboat: Ah, yes you're right. Forgot you can do that with catch2.

mmd-osm:
Thanks, looks good to me now!

mmd-osm:
I did some further refactoring in #415. Most notably, I switched back from RequestContext to request for non-HTTP POST/PUT methods, since we're not using any of the non-request fields. Previously, this looked a bit too redundant to me. 

OTOH, I have replaced the input parameters in validate_user_db_update_permission method by RequestContext, which also happens to be inside the POST/PUT code path.

Other changes are listed in the PR. I hope this makes sense.

Issue #412, opened 2024-06-19
Add the ability to limit changeset size (CGImap integration)


mmd-osm:
Tracking issue for integration of https://github.com/openstreetmap/openstreetmap-website/pull/4908 into CGImap.

We can now start working on CGImap based on this implementation. As a reminder, the code should roughly follow our previous changeset size limit approach, as explained in https://github.com/zerebubuth/openstreetmap-cgimap/pull/409#issuecomment-2171805476

New code should be implemented against the master branch and will be backported later to 0.9.x. Backporting the new code means: we will create a new branch "release/0.9.3" with the added bounding box check. This branch is based on https://github.com/zerebubuth/openstreetmap-cgimap/tree/release/0.9.2 

Related discussion: https://github.com/zerebubuth/openstreetmap-cgimap/pull/409

mmd-osm:
The database function returns a size information only, which means we would have this on CGImap as well. Started a follow up discussion in https://github.com/openstreetmap/openstreetmap-website/pull/4908 to see if we could pass {min,max}_{lat,lon} paramters to the database function instead, and the db function do the rest for us.

tomhughes:
I don't want to have to call the database functions hundreds of times though, which is why the rails code is careful to only call it once for each instance of the changeset model.

mmd-osm:
Well, we would also only call it once shortly before upload processing has finished and we're committing the changes. So from CGImap side, adding these additional four parameters would have no impact on how often we're calling the db function. Would this be different on Rails?

mmd-osm:
Before I forget it, there may also be cases where we don't have a valid bounding box at all, and the db function should  skip the size calculation.

mmd-osm:
(Please ignore previous post, if we don't have a valid bounding box, we will not call the db function at all)

tomhughes:
Sure I'll just spend another evening on this to re-engineer everything. Next week maybe?

mmd-osm:
Sure, we can go ahead with the current db function interface and revisit the topic next week.

mmd-osm:
First test in JOSM:

![image](https://github.com/zerebubuth/openstreetmap-cgimap/assets/5842757/2846f5d8-3315-4831-bd31-88f009b3d153)

talllguy:
@mmd-osm how does it look in iD for that test?

simonpoole:
1st reaction: 
[...]

PR #413, opened 2024-06-19
Add the ability to limit changeset size

mmd-osm:
See #412 for details.

For reviewers out there, please add inline comments as needed. Thanks!

Introduces a new config option: `bbox-size-limit-upload`  (boolean value)

### DIFF
--- a/include/cgimap/api06/changeset_upload/changeset_updater.hpp
+++ b/include/cgimap/api06/changeset_upload/changeset_updater.hpp
@@ -26,6 +26,8 @@ public:
 
   virtual void update_changeset(uint32_t num_new_changes, bbox_t bbox) = 0;
 
+  virtual bbox_t get_bbox() const = 0;
+
   virtual osm_changeset_id_t api_create_changeset(const std::map<std::string, std::string>&) = 0;
 
   virtual void api_update_changeset(const std::map<std::string, std::string>&) = 0;
--- a/src/api06/changeset_upload_handler.cpp
+++ b/src/api06/changeset_upload_handler.cpp
@@ -70,6 +70,26 @@ changeset_upload_responder::changeset_upload_responder(mime::type mt,
 
   changeset_updater->update_changeset(new_changes, handler.get_bbox());
 
+  if (global_settings::get_bbox_size_limiter_upload()) {
+
+    auto const cs_bbox = changeset_updater->get_bbox();
+
+    if (!(cs_bbox == bbox_t()))   // valid bbox?
+    {
+      auto const max_bbox_size = upd.get_bbox_size_limit(*user_id);
+
+      if (cs_bbox.linear_size() > max_bbox_size) {
+
+        logger::message(
+            fmt::format(
+                "Upload of {} changes by user {} in changeset {} blocked due to bbox size limit exceeded, max bbox size {}",
+                new_changes, *user_id, changeset, max_bbox_size));
+
+        throw http::payload_too_large("Changeset bounding box size limit exceeded.");
+      }
+    }
+  }
+
   upd.commit();
 }
 
--- a/include/cgimap/backend/apidb/changeset_upload/changeset_updater.hpp
+++ b/include/cgimap/backend/apidb/changeset_upload/changeset_updater.hpp
@@ -30,6 +30,8 @@ public:
 
   void update_changeset(const uint32_t num_new_changes, const bbox_t bbox) override;
 
+  bbox_t get_bbox() const override;
+
   osm_changeset_id_t api_create_changeset(const std::map<std::string, std::string>&) override;
 
   void api_update_changeset(const std::map<std::string, std::string>&) override;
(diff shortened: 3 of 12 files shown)


src/api06/changeset_upload_handler.cpp
```diff
@@ -68,6 +68,26 @@ changeset_upload_responder::changeset_upload_responder(mime::type mt,
...
 
+  if (global_settings::get_bbox_size_limiter_upload()) {
+
+    auto const cs_bbox = handler.get_bbox();
+
+    if (!(cs_bbox == bbox_t()))   // valid bbox?
+    {
+      auto const max_bbox_size = upd.get_bbox_size_limit(*user_id);
+
+      if (cs_bbox.linear_size() > max_bbox_size) {
```
Woazboat, line 79: This is only checking the bbox of the current upload. 
https://github.com/openstreetmap/openstreetmap-website/pull/4908/files#diff-33e24e983f7ae5829e0e919c3843ccaaa05be35610607ec2fd486636d2c14e56R131-R133 checks the full bounding box of the changeset after expansion.

The check needs to be moved down after `changeset_updater->update_changeset(new_changes, handler.get_bbox());` and use the expanded bounding box `changeset_updater.cs_bbox` (which is currently not accessible from outside, so a getter method is needed, or the expanded bbox should be returned from `update_changeset()`)

pnorman (approved):
I note that we're returning a HTTP 413 error in PR #413.

src/api06/changeset_upload_handler.cpp
```diff
@@ -68,6 +68,26 @@ changeset_upload_responder::changeset_upload_responder(mime::type mt,
...
+
+        logger::message(
+            fmt::format(
+                "Upload of {} changes by user {} in changeset {} blocked due to bbox size limit exceeded, max bbox size {}",
+                new_changes, *user_id, changeset, max_bbox_size));
+
+        throw http::payload_too_large("Upload has been blocked, since it exceeds the current bbox size limit.");
```
Woazboat, line 86: I'd avoid the shorthand 'bbox' in the error message since end users might not know what that means. 

"Upload has been blocked because the affected area is too large"

Woazboat (changes requested):

src/api06/changeset_upload_handler.cpp
mmd-osm: Good catch, I remember now why the previous implementation (that is no longer available) performed the check in ApiDB_Changeset_Updater::update_changeset. I think I'll move it there now.

mmd-osm: So I tested the Rails implementation with multiple uploads. The second one would fail b/c the resulting bbox is too large. I think I also need to add this scenario as a test case.

mmd-osm: This should be fixed now, I added a getter-method + added a test case.

simonpoole:
[this fits a bit better here]
1st reaction:

    we don't return 413 anywhere except for the user pref api currently (and I suspect that most editors don't use that) at least not documented and this may lead to breakage.
    the error text is too verbose, just return something like "Changeset bounding box size limit exceeded" ("blocked" has negative associations).

Will test in a couple of minutes.

include/cgimap/api06/changeset_upload/changeset_updater.hpp
```diff
@@ -26,6 +26,8 @@ class Changeset_Updater {
 
   virtual void update_changeset(uint32_t num_new_changes, bbox_t bbox) = 0;
 
+  virtual bbox_t get_bbox() = 0;
```
Woazboat, line 29: Minor nitpick: `get_bbox()` can be `const`

mmd-osm:
HTTP status code 413 is matching the Rails implementation. If there's some issue with it, it would have to be changed there as well. I'm using HTTP 413 at the moment to reject large uploads (>50MB), a feature which is not present on Rails. It's not unlikely that some editing apps have no idea how to deal with this status code.

simonpoole:
> It's not unlikely that some editing apps have no idea how to deal with this status code.

Well I assume that they will simply display the text message, but that is -very- different that its use being documented and displaying a translated message appropriate to the situation. 

Note that this is not about informing vandals, but giving unsuspecting normal users a hint they can understand what caused the problem and what is the best action to take.

mmd-osm:
I'm going to merge this now, to open up testing for everyone on the dev instance https://master.apis.dev.openstreetmap.org

Once it is merged, can you (@tomhughes) please add a new config setting in https://github.com/openstreetmap/chef/blob/master/cookbooks/dev/templates/default/cgimap.environment.erb 

```
CGIMAP_BBOX_SIZE_LIMIT_UPLOAD="true"
```

for this new parameter?

```
  --bbox-size-limit-upload arg  enable bbox size limit for changeset upload
```

PR #428, opened 2024-07-17
Add benchmark framework

Woazboat:
Add a benchmarking framework using the https://github.com/google/benchmark library.

Could be useful to judge the impact of changes like https://github.com/zerebubuth/openstreetmap-cgimap/pull/427 or for finding performance regressions.

Not really polished yet and this PR is mainly intended for discussion if this is wanted or not. The included benchmarks are also mostly just to be used as an example. 

Usage:
`pg_virtualenv ./cgimap-benchmark`
or e.g.
`pg_virtualenv ./cgimap-benchmar --benchmark_out=bm_result.json --benchmark_out_format=json --benchmark_repetitions=5`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,7 @@ option(ENABLE_COVERAGE "Compile with coverage info collection" OFF)
 option(ENABLE_PROFILING "Compile with profiling" OFF)
 option(BUILD_SHARED_LIBS "Build using shared libraries" OFF)
 option(ENABLE_INSTALL "Enable installation" ON)
+option(BUILD_BENCHMARK "Compile benchmarks" OFF)
 if(CMAKE_BUILD_TYPE STREQUAL "Debug")
     option(BUILD_TESTING "Compile test suite" ON)
 else()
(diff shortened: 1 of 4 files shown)


mmd-osm:
I think google benchmark could be a good fit for cpu bounded activities, like parsing JSON/XML or generating XML/JSON responses, maybe also including different compression settings. We could define different scenarios with varying number of nodes/ways/relations and create/modify/delete operations, as needed. Overall, this would be totally fine and a good enhancement.

With database queries, we have a lot more boundary conditions, like the database version, any parameter settings, storage system, up-to-date indexes, existing data, test data volume, etc. In particular the last two points are a bit tricky, because queries could be fast on a nearly empty database, and very slow once we reach several 100 million entries (see OHM experience). They could be slow due to some optimizer decision, or due to some poor query design. The next scenario is handling parallel requests (like multiple users uploading changes at the same time). Also here we need to make sure that we're properly scaling across requests, that users are not locking each other, etc. 

Long story short, I'm not convinced that google-benchmark is the right tool for this use case. I'm currently using different command line tools to simulate single/multiple users as well as existing editing apps for testing, and some off the shelf tools for db analysis.

PR #431, opened 2024-07-26
std::from_chars for number parsing

mmd-osm:


include/cgimap/api06/changeset_upload/node.hpp
```diff
@@ -31,32 +30,30 @@ class Node : public OSMObject {
...
+    auto [_, ec] = std::from_chars(lat.data(), lat.data() + lat.size(), _lat);
+
+    if (ec == std::errc())
+      set_lat(_lat);
+    else if (ec == std::errc::invalid_argument)
       throw xml_error("Latitude is not numeric");
-    } catch (std::out_of_range &e) {
+    else if (ec == std::errc::result_out_of_range)
       throw xml_error("Latitude value is too large");
-    }
```
Woazboat, line 42: The exception catch block was exhaustive and would bubble up uncaught exception types. The if checks are no longer exhaustive and will simply fall through without an error being raised if an unmatched value is returned.

I know https://en.cppreference.com/w/cpp/utility/from_chars only mentions `std::errc::invalid_argument` and `std::errc::result_out_of_range`, but I'm not sure if that is a guarantee or how future proof that is. Maybe a catch all `else` could be added to be on the safe side?

mmd-osm: I also thought about this. Except for `add_way_node` we're using `std::optional` for all other values. These would trigger an error later on in `is_valid()`, if they happen to be empty.

Adding a catch all `else` for sure wouldn't hurt here.

mmd-osm: This should be covered in #432 now.

PR #447, opened 2024-08-15
Debian package build

mmd-osm:
This PR is based on work by @Woazboat  in d499553211747d8df75cd6647d04769e03aa0c89, with some minor tweaks to upload the generated Debian build artifacts later on. Also, Debian 12 Dockerfile has additional compile flags similar to Ubuntu launchpad config.

### DIFF
--- a/.github/workflows/docker_bookworm.yml
+++ b/.github/workflows/docker_bookworm.yml
@@ -24,4 +24,10 @@ jobs:
 
       - name: Running Docker image
         run: |
-           docker run --entrypoint /bin/bash cgimap:bookworm -c "/usr/local/bin/openstreetmap-cgimap --help"
+           docker run --name cgimap --entrypoint /bin/bash cgimap:bookworm -c "/usr/bin/openstreetmap-cgimap --help"
+           docker cp cgimap:/app_deb ${{ github.workspace }}
+      - uses: actions/upload-artifact@v4
+        with:
+          name: openstreetmap_cgimap_debian_12.zip
+          path: ${{ github.workspace }}/app_deb/*.deb
+          if-no-files-found: error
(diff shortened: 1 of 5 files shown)


Firefishy:
I am working on another route for building debian packages over here: https://github.com/Firefishy/openstreetmap-cgimap/tree/add-debian-build

mmd-osm:
Thank you for checking some alternatives. Did you check out https://gitlab.kitware.com/debian/dh-cmake already? This might come in handy to further streamline the build, since it can leverage the cmake dependency graph. Right now, it seems you would define all dependencies manually. I read somewhere that this would be needed for "official" Debian packages only (cpack debian files are not fully compliant). For our purposes this most likely wouldn't matter anyway.

mmd-osm:
So the current list of dependencies for Debian 12 should be:

` Depends: libboost-program-options1.74.0 (>= 1.74.0+ds1), libbrotli1 (>= 0.6.0), libc6 (>= 2.34), libcrypto++8 (>= 8.7.0+git220824), libfcgi0ldbl (>= 2.4.2), libgcc-s1 (>= 3.3.1), libmemcached11 (>= 1.1.4), libpqxx-6.4 (>= 6.4), libstdc++6 (>= 11), libxml2 (>= 2.7.4), libyajl2 (>= 2.0.4), zlib1g (>= 1:1.1.4)`

That's pretty much what cpack came up with.

mmd-osm:
@Firefishy : for better visibility, you could also create a new issue and post any findings there.

mmd-osm:
@tomhughes mentioned he's urgently looking for a new build. Should we wait for this new approach to be ready first? I may have some limited availability in the next few weeks...

tomhughes:
I mean it's not critical or anything it's just moving to Debian will get us ruby 3.1 which unblocks rails 7.2 for us.

PR #454, opened 2024-09-08
Adjustments for libpqxx-7.x, clang, macOS

mmd-osm:
Without "inline", clang 17 was rightfully complaining about multiple definitions of `pqxx::type_name<... >'

### DIFF
--- a/include/cgimap/backend/apidb/pqxx_string_traits.hpp
+++ b/include/cgimap/backend/apidb/pqxx_string_traits.hpp
@@ -127,7 +127,6 @@ PQXX_ARRAY_STRING_TRAITS(std::vector<tile_id_t>);
 PQXX_ARRAY_STRING_TRAITS(std::vector<osm_changeset_id_t>);
 PQXX_ARRAY_STRING_TRAITS(std::set<osm_changeset_id_t>);
 PQXX_ARRAY_STRING_TRAITS(std::vector<std::string>);
-PQXX_ARRAY_STRING_TRAITS(std::vector<bool>);
 
 } // namespace pqxx
 
--- a/include/cgimap/backend/apidb/pqxx_string_traits.hpp
+++ b/include/cgimap/backend/apidb/pqxx_string_traits.hpp
@@ -116,7 +116,7 @@ public:
   template<>                                                            \
   struct string_traits<type> : cgimap::array_string_traits<type> {};    \
                                                                         \
-  template<> std::string const type_name<type>{#type};                  \
+  template<> inline std::string const type_name<type>{#type};           \
 
 
 #endif
--- a/src/backend/apidb/changeset_upload/relation_updater.cpp
+++ b/src/backend/apidb/changeset_upload/relation_updater.cpp
@@ -1239,17 +1239,16 @@ void ApiDB_Relation_Updater::update_current_relations(
 
   m.prepare("update_current_relations",
             R"(   
-        WITH u(id, changeset_id, visible, version) AS (
+        WITH u(id, changeset_id, version) AS (
                 SELECT * FROM
                 UNNEST( CAST($1 AS bigint[]),
                         CAST($2 AS bigint[]),
-                        CAST($3 AS boolean[]),
-                        CAST($4 AS bigint[])
+                        CAST($3 AS bigint[])
                       )
         )
         UPDATE current_relations AS r
         SET changeset_id = u.changeset_id,
-            visible = u.visible,
+            visible = CAST($4 as boolean),
             timestamp = (now() at time zone 'utc'),
             version = u.version + 1
         FROM u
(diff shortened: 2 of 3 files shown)


mmd-osm:
Ping @deevroman : this may have been triggering your issue on macOS.

deevroman:
> Ping @deevroman : this may have been triggering your issue on macOS.

Hmm, I don't think I've encountered this error.  (I compile with Clang 15) 

Just in case: before getting to the linking stage, I had to quickly fix several pieces of code. For example, Apple Clang does not have an implementation of std::from_chars

via https://en.osm.town/@foxy/113100850337066645

mmd-osm:
Yes, the issue reported by the linker is slightly different. This still needs to be fixed nevertheless.

mmd-osm:
Currently, we're using std::vector<bool> for visible flags in the following locations:

* https://github.com/zerebubuth/openstreetmap-cgimap/blob/master/src/backend/apidb/changeset_upload/relation_updater.cpp#L1263
* https://github.com/zerebubuth/openstreetmap-cgimap/blob/master/src/backend/apidb/changeset_upload/way_updater.cpp#L666

Woazboat:
Can libpqxx deal with using integral types as booleans? If yes we should probably change that to e.g. `std::vector<char>`/`std::vector<uint8_t>`. That should be both faster and less problematic.

mmd-osm:
Well, if we move the cast to bool  inside the SQL statement, we could use something else than vector<bool>.

mmd-osm:
Well, we can probably take advantage of the "cast to bool[]"  inside the SQL statement, and pass  ` {'true', 'false'} ` instead of `vector<bool>` to the query. Need to give this a try, if it works out of the box already.

Woazboat:
`0`/`1` is also fine in the input according to the postgresql documentation, so just changing the vector type would probably work(?)
https://www.postgresql.org/docs/current/datatype-boolean.html

mmd-osm:
Yes, that would work. I just noticed that we don't need a "vector or anything" at all. The value is taken from a method input parameter and never changes. The following would do as well:

```
  m.prepare("update_current_ways", R"(
      WITH u(id, changeset_id, version) AS (
         SELECT * FROM
         UNNEST( CAST($1 AS bigint[]),
                 CAST($2 AS bigint[]),
                 CAST($3 AS bigint[])
              )
      )
      UPDATE current_ways AS w
      SET changeset_id = u.changeset_id,
         visible = CAST($4 as boolean),
         timestamp = (now() at time zone 'utc'),
         version = u.version + 1
         FROM u
      WHERE w.id = u.id
      AND   w.version = u.version
      RETURNING w.id, w.version
     )");

  std::vector<osm_nwr_signed_id_t> ids;
  std::vector<osm_changeset_id_t> cs;
  std::vector<osm_version_t> versions;
  std::map<osm_nwr_id_t, osm_nwr_signed_id_t> id_to_old_id;

  for (const auto &way : ways) {
    ids.emplace_back(way.id);
    cs.emplace_back(way.changeset_id);
    versions.emplace_back(way.version);
    id_to_old_id[way.id] = way.old_id;
  }

  pqxx::result r =
      m.exec_prepared("update_current_ways", ids, cs, versions, visible);

```

deevroman:
> Can libpqxx deal with using integral types as booleans? If yes we should probably change that to e.g. `std::vector<char>`/`std::vector<uint8_t>`. That should be both faster and cause fewer issues.

I can say that after replacing vector<bool> with vector<int> the linking is successful!

replacement with vector<int8_t> ends with

```
Undefined symbols for architecture arm64:
  "void pqxx::oops_forbidden_conversion<signed char>()", referenced from:
      pqxx::forbidden_conversion<signed char>::size_buffer(signed char const&) in libcgimap_apidb.a[12](relation_updater.cpp.o)
      pqxx::forbidden_conversion<signed char>::into_buf(char*, char*, signed char const&) in libcgimap_apidb.a[12](relation_updater.cpp.o)
ld: symbol(s) not found for architecture arm64
```

mmd-osm:
I pushed another commit to get rid of `std::vector<bool>` altogether in the two methods I mentioned earlier on.

mmd-osm:
@deevroman : maybe you could re-test the latest master, and see if you encounter any issues. I don't expect any linker errors, but you never know without testing...

deevroman:
I tested it, there are no more errors from the linker 👍 A little later I’ll collect the places in the code that I corrected so that the project starts compiling

mmd-osm:
Thanks. I think it would be helpful to see the error messages for the unmodified code. Maybe some additional compile time flags can fix some of the issues without changing the code (example: https://stackoverflow.com/questions/76932735/unable-to-use-to-chars-with-clang-on-macos).

PR #461, opened 2025-01-04
Diff upload: minor improvements

mmd-osm:
See single commits for details

### DIFF
--- a/src/backend/apidb/readonly_pgsql_selection.cpp
+++ b/src/backend/apidb/readonly_pgsql_selection.cpp
@@ -786,8 +786,8 @@ bool readonly_pgsql_selection::supports_user_details() const {
 bool readonly_pgsql_selection::is_user_blocked(const osm_user_id_t id) {
 
   m.prepare("check_user_blocked",
-    R"(SELECT id FROM "user_blocks" 
-          WHERE "user_blocks"."user_id" = $1 
+    R"(SELECT id FROM "user_blocks"
+          WHERE "user_blocks"."user_id" = $1
             AND (needs_view or ends_at > (now() at time zone 'utc')) LIMIT 1 )");
 
   auto res = m.exec_prepared("check_user_blocked", id);
@@ -826,7 +826,7 @@ std::optional< osm_user_id_t > readonly_pgsql_selection::get_user_id_for_oauth2_
   m.prepare("oauth2_access_token",
     R"(SELECT resource_owner_id as user_id,
          CASE WHEN expires_in IS NULL THEN false
-              ELSE (created_at + expires_in * interval '1' second)  < now() at time zone 'utc'
+              ELSE (created_at + expires_in * interval '1' second) < now() at time zone 'utc'
          END as expired,
          COALESCE(revoked_at < now() at time zone 'utc', false) as revoked,
          'write_api' = any(string_to_array(scopes, ' ')) as allow_api_write
--- a/src/backend/apidb/readonly_pgsql_selection.cpp
+++ b/src/backend/apidb/readonly_pgsql_selection.cpp
@@ -91,19 +91,18 @@ osm_edition_t id_of<osm_edition_t>(const pqxx_tuple &row, pqxx::row_size_type co
 
 template <typename T>
 inline int insert_results(const pqxx::result &res, set<T> &elems) {
-  int num_inserted = 0;
 
   auto const id_col = res.column_number("id");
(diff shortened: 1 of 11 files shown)


src/backend/apidb/readonly_pgsql_selection.cpp
```diff
@@ -91,19 +91,18 @@ osm_edition_t id_of<osm_edition_t>(const pqxx_tuple &row, pqxx::row_size_type co
...
+
+  auto it = elems.begin();
   for (const auto & row : res) {
     const T id = id_of<T>(row, id_col);
-
-    // note: only count the *new* rows inserted.
-    if (elems.insert(id).second) {
-      ++num_inserted;
-    }
+    it = elems.emplace_hint(it, id);
```
Woazboat, line 102: I think the hint iterator position here is never correct at the moment. 

`emplace`/`insert` returns an iterator to the inserted element itself (/the existing element with the same value). 
The hint iterator needs to point to the element _after_ the position where the new element should be inserted. (i.e. the new element will be inserted _before_ the iterator). 

That means the next new element would have to be _lower_ than the previous one in order for the hint to be correct if the iterator returned from `emplace`/`insert` is directly used as a hint. 
Advancing the returned iterator one step so that it points past the newly inserted element should fix the problem.

mmd-osm: The new code should follow exactly the pattern as in `set_emplace_hint_closest` from https://en.cppreference.com/w/cpp/container/set/emplace_hint, there's also no emplace or insert anymore, only the emplace_hint. It also looked ok in the perf trace...

Woazboat: I think the example code might actually be wrong.
https://quick-bench.com/q/mPl0Ufd9dkzbAhClYurznUjzcFE

 Insertion _after_ the hint iterator was the pre C++11 behavior for `insert()` with a position hint, but that was changed to insertion _before_ the hint iterator with C++11.
https://en.cppreference.com/w/cpp/container/set/insert
> Complexity
...
3,4) Amortized constant if the insertion happens in the position just after(until C++11)before(since C++11) pos, logarithmic in the size of the container otherwise.

I used `emplace`/`insert` as stand-ins for the whole family of insertion functions since they all have the same behavior.

mmd-osm: Here's how this looked like in my perf trace. The code snippet is taken from _M_get_insert_hint_unique_pos in stl_tree.h. This code first tries to locate a matching position before and then afterwards the hint position.

Since each new value is larger than all previous values (for the very first time we're calling insert_results), the hint returns the rightmost position. I don't see any indication of a binary search via  _M_get_insert_unique_pos in the trace.

image

Woazboat: I think the implementation might just try both before and after positions for the hint. That's implementation defined behavior though and not required by the standard.

mmd-osm: I have added another test case to the cppreference sample code: https://gist.github.com/mmd-osm/f7a7665738e9f564b25ab05da1cd3be6 

For some reason, the additional `++it ` has a fairly large impact on the overall runtime. It's not clear to me why this isn't showing up in the quickbench. Test on gcc 12.

```
278.23ms for plain emplace (ratio: 1.00)
 56.06ms for emplace with correct hint (ratio: 4.96)
281.64ms for emplace with wrong hint (ratio: 0.99)
 44.64ms for corrected emplace (ratio: 6.23)
 47.24ms for emplace using returned iterator (ratio: 5.89)
236.48ms for emplace using returned iterator_increment (ratio: 1.18)
```

image

Woazboat: Hmmm, apparently the external iterator increment is much more expensive than the lookahead to the next position it does internally for whatever reason. It does require an additional element comparison, but that's really cheap when you only have integers.

I guess if it works, it works...

mmd-osm: I think in case we start with an empty set,  `_M_get_insert_hint_unique_pos` can avoid incrementing the iterator altogether by comparing the hint element with the M_rightmost element (see screenshot https://github.com/zerebubuth/openstreetmap-cgimap/pull/461#discussion_r1903282347 above).

PR #466, opened 2025-02-23
New AcceptHeader parser w/o boost

mmd-osm:


src/choose_formatter.cpp
```diff
@@ -28,233 +29,190 @@
...
-      mapping.insert(make_pair(mime_type, quality));
-    }
+  for (const auto &acceptedType : acceptedTypes)
+    mapping[acceptedType.mimeType] = acceptedType.q;
 
-  } else {
-    logger::message(fmt::format("Failed to parse accept header '{}'",
-                    accept_header));
+  if (mapping.empty()) {
     throw http::bad_request("Accept header could not be parsed.");
```
Woazboat, line 50: This should be a `406 Not Acceptable` instead of a `400 Bad Request` if there are no accepted media types in the header. (Non-accepted types are currently simply skipped in `parse()`, which can cause the map to be empty)

Not sure if the constructor should be throwing exceptions at all though unless the input string is malformed?

src/choose_formatter.cpp
```diff
@@ -28,233 +29,190 @@
...
+    // parameters
+    auto elems = split_trim(item, ';');
+
+    if (elems.empty()) {
+      continue;
+    }
+
+    // Treat Accept: * as Accept: */*
+    if (elems[0] == "*")
+      elems[0] = "*/*";
```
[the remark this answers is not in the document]
mmd-osm: Yup, there’s this funny unit test: https://github.com/zerebubuth/openstreetmap-cgimap/blob/1f91c1630da24040c2f9bd72f5aa62dcba1c38ca/test/empty.testcore/accept_any.case#L3

include/cgimap/util.hpp
```diff
@@ -90,6 +90,43 @@ inline bool iequals(std::string_view a, std::string_view b) {
...
+// Remove leading and trailing whitespace from string
+inline std::string_view trim(std::string_view str) {
+  auto start = str.find_first_not_of(" \t\n\r");
+  if (start == std::string::npos)
+    return {};
+  auto end = str.find_last_not_of(" \t\n\r");
+  return str.substr(start, end - start + 1);
+}
+
+inline std::vector<std::string_view> split_trim(std::string_view str, char delim) {
```
mmd-osm, line 103: This could be replaced later on by a c++20 version: https://github.com/mmd-osm/openstreetmap-cgimap/blob/cpp20_1/include/cgimap/util.hpp#L104-L117

Woazboat:
I've made some changes to this here: https://github.com/Woazboat/openstreetmap-cgimap/commit/b84c1f1d9124b22a4c09a13af4ee9f4914f2dde3

Parsing the q param value correctly according to the spec is a problem. [std::stod()](https://en.cppreference.com/w/cpp/string/basic_string/stof) accepts all sorts of input values that are not allowed by the spec, such as float values with exponents or hexadecimal numbers. It's also locale dependent, which can change the decimal point character and break the parsing.

[std::from_chars](https://en.cppreference.com/w/cpp/utility/from_chars) would solve those problems, but then we run into the issue again that it's not implemented for floating point types on apple platforms.

mmd-osm:
(rebase on master)

mmd-osm:
node.hpp already has some special logic for Apple to parse double values.

PR #473, opened 2025-03-02
Enable unindent by checking compiler flag instead

mmd-osm:


include/cgimap/backend/apidb/utils.hpp
```diff
@@ -19,8 +19,9 @@
 
 /* Unindent SQL statements in raw strings */
 
-#if __GNUC__ < 12
+#if !defined(__cpp_lib_constexpr_vector)
```
Woazboat, line 22: This should be `#if __cpp_lib_constexpr_vector`. 
The macro could be defined (or intentionally overridden) as `0` when the feature is not supported, which would still evaluate to true when just checking whether it is defined.
Undefined macros will always evaluate to `0` inside preprocessor `#if` conditions and the feature check macros are guaranteed to be non-zero if the feature is implemented.

Issue #469, opened 2025-02-28
Replacing fmt lib with std::format


Woazboat:
Creating this issue mostly for documenting the challenges involved, since it came up as a potential change after switching to C++20.

Obviously depends on at least C++20 https://github.com/zerebubuth/openstreetmap-cgimap/issues/323

- Replacing instances of `fmt::format` is trivial and just requires a rename to `std::format`
- `FMT_COMPILE` macros for optimized fmt format strings can simply be removed. `std::format` is already compile time.
- Some code snippets currently choose between different more optimized implementations depending on the version of the fmt lib. These could be simplified and replaced by the equivalent functions which are guaranteed to exist in std `<format>`.
- Custom `fmt::formatter` helpers (for e.g. the `api06::id_version` type) need to be rewritten to the [std::formatter](https://en.cppreference.com/w/cpp/utility/format/formatter)  structure required by the stdlib.
- `fmt::join()` is currently used in several places to format ranges with separators (`"1,2,3,4"`). There is unfortunately no direct analogue available in C++20. Range formatting was not included in C++20 and was only introduced in [C++23](https://en.cppreference.com/w/cpp/utility/format/range_formatter). It is also [not yet supported by GCC libstdc++](https://en.cppreference.com/w/cpp/23).
- `std::format` requires [at least GCC 13](https://en.cppreference.com/w/cpp/20), Debian stable [still ships GCC 12](https://packages.debian.org/search?keywords=g%2B%2B)

Woazboat:
https://github.com/Woazboat/openstreetmap-cgimap/tree/c%2B%2B20_replace_fmt has all changes except a replacement for fmt:::join

mmd-osm:
A simplified replacement for fmt::join() might be good enough to format a few ids: https://cpp.godbolt.org/z/eovh9re8v

Woazboat:
PR https://github.com/zerebubuth/openstreetmap-cgimap/pull/474 should be revisited when support for GCC 12 is no longer required (see initial comment).

PR #474, opened 2025-03-03
Replace fmt lib with C++20 std::format

Woazboat:
Merging is not possible yet because `std::format` requires GCC 13, which is not available in current Debian stable.

Closes https://github.com/zerebubuth/openstreetmap-cgimap/issues/469


mmd-osm:
What do you think about this polyfill approach? https://stackoverflow.com/a/66989192

Woazboat:
I don't think it's worth it. The main benefit of using std::format over fmt lib is a minor reduction of dependencies and the associated complexity. Trying to maintain both simultaneously would only make things worse in that aspect. 

As far as dependencies go, fmt lib is very unproblematic. I think keeping it around a little longer is totally fine.

Woazboat:
The question is, should this just be closed and reopened when it's possible to do so or should it kick around as a permanent draft PR?

mmd-osm:
I will at least merge the unit tests in https://github.com/zerebubuth/openstreetmap-cgimap/pull/474/commits/0291ebaa95095e4cf35930875c8cfd2687d60014 in another PR (with some small changes).

Maybe we can leave a comment in #469 that this PR should be revisited at some point in the future, and then only keep #469 open?

PR #477, opened 2025-03-04
Util unit tests from #474

mmd-osm:
Test for is_comma_separated_string is part of #474 and not yet merged

Edit: we're testing to_string instead now.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -92,6 +92,22 @@ if(BUILD_TESTING)
         COMMAND test_http)
 
 
+    ###########
+    # test_utils
+    ###########
+    add_executable(test_utils
+        test_utils.cpp)
+
+    target_link_libraries(test_utils
+        cgimap_common_compiler_options
+        cgimap_core
+        Boost::program_options
+        catch2)
+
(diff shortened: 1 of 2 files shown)


Woazboat:
The `is_comma_separated_string()` function is just `to_string()` from util.hpp renamed and changed to avoid fmt lib.

mmd-osm:
Good point. We don't seem to have any unit test covering "The following ... ids are not known on the database: ...". That's about the only place where this helper function is called. So let's add it back with the right name.

Woazboat:
Btw., is `CHECK(parse_ruby_number<uint32_t>("123abc") == 123);` the intended behavior or should it rather fail when it encounters invalid characters?

mmd-osm:
I think it's fine the way it is now. I didn't remember how to_i is implemented in Ruby and had to give it a try on the rails console:

```
"123abc".to_i
=> 123
```

We seem to have some deviations for very large or negative numbers... I think that's fine for the parse_single_id use case and shouldn't really matter.

```
"9999999999999999999999999".to_i
=> 9999999999999999999999999
```

```
"-1".to_i
=> -1
```

PR #481, opened 2025-03-06
 Simplified (unit test) code with c++20 /2

mmd-osm:


include/cgimap/util.hpp
```diff
@@ -93,35 +93,40 @@ inline bool iequals(std::string_view a, std::string_view b) {
          std::ranges::equal(a, b, ichar_equals);
 }
 
-inline std::string_view trim(std::string_view str) {
+template <typename T>
+concept StringLike = std::is_same_v<T, std::string> || std::is_same_v<T, std::string_view>;
```
Woazboat, line 97: I _think_ this might need https://en.cppreference.com/w/cpp/types/remove_cvref to avoid weird interactions with const/references.

mmd-osm: This seems plausible. I added it for now... will need to take a closer look.

mmd-osm: By the way, I only added the concept / template parameter to avoid repeated string -> string_view -> string conversions in places like  `auto key = std::string(trim(line_view.substr(0, pos)));`

PR #536, opened 2025-06-23
Update Catch2 unit test framework v2.13.10 -> v3.8.1

Woazboat:
- Update bundled Catch2 library + fix breaking changes
- Add build option to use system Catch2 library instead of the one included in the repo
- Only build Catch2 lib if required (i.e. for debug builds with tests)


mmd-osm:
Thanks a lot for looking into the Catch2 update, well done! Coincidentally, I briefly looked into the topic a few months ago and thought that it would be too much effort to migrate all unit tests... 

On Debian Trixie, I gave the libcatch2-dev system library a try, and this worked without issues.

Woazboat:
The only thing that really needed to be changed in the cgimap test code were some of the class namespaces and that was a simple matter of starting a build, seeing what breaks, do a find + replace with the new namespace, repeat until everything compiles. Changing the header imports to the piecewise headers was pretty similar (and technically entirely optional).

The far more annoying part was modifying the imported catch2 subtree to remove all the unnecessary files in a way that they never even show up in this git repo so that they don't increase the repo size.

PR #542, opened 2026-03-06
Refactor HTTP Accept-Encoding header parsing

### LINKED ISSUE #540: Accept-Encoding is not parsed correctly
The current parsing of the `Accept-Encoding` header requires `, ` exactly as the delimiter:

https://github.com/zerebubuth/openstreetmap-cgimap/blob/01e33669bfee17d55849d6f731a174b8058ca630/src/http.cpp#L161-L163

In reality `,` is the delimiter and it [allows optional whitespace before and after delimiters](https://httpwg.org/specs/rfc9110.html#abnf.extension).

There is a similar problem with quality values:

https://github.com/zerebubuth/openstreetmap-cgimap/blob/01e33669bfee17d55849d6f731a174b8058ca630/src/http.cpp#L181-L183

which [should allow optional whitespace either side of the semicolon](https://httpwg.org/specs/rfc9110.html#rfc.section.12.4.2).

Woazboat:
Implement a generic HTTP list header parser and use it to replace the current Accept-Encoding header parsing. 

Properly handle optional whitespace in list items/parameters. Prioritize encodings in order requested by user (https://github.com/zerebubuth/openstreetmap-cgimap/issues/540#issuecomment-3554307304).

Change encoding classes to singletons since they are only used as factories and don't hold any state.

---

Requires GCC version 12 to make `std::ranges::views::split` actually usable 
- C++20 defect report [P2210R2](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2210r2.html)
- https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html

GCC 12 is [available](https://documentation.ubuntu.com/ubuntu-for-developers/reference/availability/gcc/) in Ubuntu 22.04 used in CI, but GCC 11 is still used as the default 

(GCC 12 is also required for constexpr `std::string`/`std::vector`, but those aren't really necessary.)

https://en.cppreference.com/w/cpp/20.html

---

Fixes https://github.com/zerebubuth/openstreetmap-cgimap/issues/540

### DIFF
--- a/include/cgimap/api06/changeset_upload/osmobject.hpp
+++ b/include/cgimap/api06/changeset_upload/osmobject.hpp
@@ -12,6 +12,7 @@
 
 #include "cgimap/types.hpp"
 #include "cgimap/util.hpp"
+#include "cgimap/http.hpp"
 
 #include <charconv>
 #include <map>
--- a/include/cgimap/http.hpp
+++ b/include/cgimap/http.hpp
@@ -339,7 +339,7 @@ public:
  * Parses an Accept-Encoding header and returns the chosen
  * encoding.
  */
-const http::encoding* choose_encoding(const std::string &accept_encoding);
+const http::encoding* choose_encoding(std::string_view accept_encoding);
 
 std::unique_ptr<ZLibBaseDecompressor> get_content_encoding_handler(std::string_view content_encoding);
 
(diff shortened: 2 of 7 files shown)


mmd-osm:
OWG mitigated this issue back in 2025 by defining appropriate Apache rewrite rules, after the previous CDN had been discontinued on very short notice. A bit later requests were handled by Fastly, where the issue didn't show up anymore to my knowledge.

So overall this has low priority - no immediate impact on production.

PR #543, opened 2026-06-11
Include libxml/xmlerror.h for xmlSetGenericErrorFunc

tomhughes:
The build was failing with libxml2 2.12.10 on Fedora 44 because `xmlSetGenericErrorFunc` was not defined.

I assume older versions must have been pulling in `xmlerror.h` as a side effect of one the other headers we include but we should always really have been including it explicitly.

### DIFF
--- a/src/xml_writer.cpp
+++ b/src/xml_writer.cpp
@@ -15,6 +15,8 @@
 #include <fmt/core.h>
 #include <fmt/compile.h>
 
+#include <libxml/xmlerror.h>
+
 // C-style callback functions are called from xmlOutputBufferCreateIO.
 // Functions cannot throw, return -1 if an error occurred.
 static int wrap_write(void *context, const char *buffer, int len) noexcept {


Woazboat:
The arch CI build also fails because of the missing include.

Woazboat:
Ping @mmd-osm , are you around?

Firefishy:
@zerebubuth Any chance you could look at this?

zerebubuth:
Do you need me to tag a new version too?

Firefishy:
Appreciated, I don't think there is an urgent need for a new version. There was some community fear that cgimap may no longer have any maintainers. All good now :-)

mmd-osm:
Hi folks, I'm fairly busy right now and can only look into issues, in case they are important, urgent, production related without workaround available. I'm afraid that everything else needs to wait at the moment. This fix didn't seem to meet any of these criteria, so I just let it sit there. Take care.

zerebubuth:
@mmd-osm no worries, hope everything is okay. Feel free to @ mention me if there are things that fall in the gap between urgent and backlog. I can't guarantee I'll be able to help, but I will if I can.

